PR #12, opened 2023-02-20
Unit tests and image scanner

bencekov:
This pr includes the currently unfinished unit tests, and the updated github workflow with container image scanning with trivy.


massimilianogo:
@bencekov please split the build.yaml workflow in smaller workloads which perform one action only (e.g. build, test, scan, publish)
You can refer to [this PR](https://github.com/canonical/nginx-rock/pull/4) as an example.

main_test.go
```diff
@@ -0,0 +1,156 @@
+package main
```
gruyaume, line 1: I don't see the place in the CI where those are executed, please make sure they at least run on every PR.

main_test.go
```diff
@@ -0,0 +1,156 @@
...
+	s := httptest.NewServer(mux)
+	os.Setenv("HYDRA_ADMIN_URL", s.URL+"/")
+	return s
+}
+
+// kratos GET //self-service/login/browser?aal=&login_challenge=&refresh=false&return_to= HTTP/1.1
+// hydra Put /oauth2/auth/requests/login/accept
+func TestHandleCreateFlow(t *testing.T) {
+	kratosClient := CreateKratosMockServer()
+	hydraClient := CreateHydraMockServer()
```
nsklikas, line 38: This will happen in every test, let's move it to a `setup` function.

nsklikas (changes requested):
Good job. I have a couple of comments:
- The tests need to check that the serve behaves as expected when errors occur as well. For every endpoint we need to check what happens if we get an unexpected response from kratos/hydra. 
- We need to test what will happen if the server gets a bad/malformed parameter in the request. 
- Are you going to implement the rest of the server mock responses in this PR or is it going to be done in another one?

PR #13, opened 2023-02-24
Unit tests and github workflow

bencekov:
Apologies for multiple changes in one pr. This pr includes the following:
- Finished (for now) unit tests
- github workflow separated into multiple files
-  added unit tests to build workflow


main_test.go
```diff
@@ -0,0 +1,291 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// --------------------------------------------
+// TESTING WITH CORRECT SERVERS
+// --------------------------------------------
+func TestHandleCreateFlowWithoutCookie(t *testing.T) {
```
nsklikas, line 21: `testing.T` provides `Cleanup` function (https://pkg.go.dev/testing#T.Cleanup). You could pass it to `testServers.CreateTestServers` and the servers can then register their own clean up methods so that you don't have to do it in the test.

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+	if res.StatusCode != 200 {
+		t.Errorf("expected StatusCode to be 200 got %v", res.StatusCode)
+	}
+	requestLoginResponse := new(testLoginBrowser.OAuth2RequestLoginResponse)
+	if err := json.Unmarshal(data, requestLoginResponse); err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	if requestLoginResponse.Redirect_to != "test.test" {
+		t.Errorf("expected test.test, got %v", string(data))
+	}
```
nsklikas, line 81: You should use [assertions](https://pkg.go.dev/github.com/stretchr/testify/assert) to verify the response's output

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	if res.StatusCode != 200 {
+		t.Errorf("expected StatusCode to be 200 got %v", res.StatusCode)
+	}
+	requestLoginResponse := new(testLoginBrowser.OAuth2RequestLoginResponse)
+	if err := json.Unmarshal(data, requestLoginResponse); err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
```
nsklikas, line 78: This is repeated in too many places, perhaps we could move this logic to a `UnamarshalJsonBody` (or something like that) function.

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+// currently only prints out results main.go needs pr to handle timeouts
+// --------------------------------------------
+func TestHandleCreateFlowTimeout(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateTimeoutServers, http.MethodPut,
+		"/api/kratos/self-service/login/browser?aal=aal1&login_challenge=&refresh=false&return_to=http://test.test",
+		nil, handleCreateFlow)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 185: We are not asserting something in those tests. We don't gracefully handle server errors in the server yet, but if the Kratos/Hydra request times out we should just return a `500` HTTP response with some explanatory error message (for now).

Let's leave a comment about it (like `TODO: Validate response when server error handling is implemented) here.

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	bodyReader := bytes.NewReader(bodyJson)
+	data, err := CreateGenericTest(testServers.CreateTimeoutServers, http.MethodPost,
+		"/api/kratos/self-service/login?flow=1111",
+		bodyReader, handleUpdateFlow)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 204: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+}
+func TestHandleKratosErrorTimeout(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateTimeoutServers, http.MethodGet,
+		"/api/kratos/self-service/errors?id=1111",
+		nil, handleKratosError)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 213: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+}
+func TestHandleConsentTimeout(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateTimeoutServers, http.MethodGet,
+		"/api/consent?consent_challenge=test_challange",
+		nil, handleConsent)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 222: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+// currently only prints out results main.go needs pr to handle errors
+// --------------------------------------------
+func TestHandleCreateFlowError(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateErrorServers, http.MethodPut,
+		"/api/kratos/self-service/login/browser?aal=aal1&login_challenge=&refresh=false&return_to=http://test.test",
+		nil, handleCreateFlow)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 236: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	bodyReader := bytes.NewReader(bodyJson)
+	data, err := CreateGenericTest(testServers.CreateErrorServers, http.MethodPost,
+		"/api/kratos/self-service/login?flow=1111",
+		bodyReader, handleUpdateFlow)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 255: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+}
+func TestHandleKratosErrorError(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateErrorServers, http.MethodGet,
+		"/api/kratos/self-service/errors?id=1111",
+		nil, handleKratosError)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 264: Same

main_test.go
```diff
@@ -0,0 +1,291 @@
...
+}
+func TestHandleConsentError(t *testing.T) {
+	data, err := CreateGenericTest(testServers.CreateErrorServers, http.MethodGet,
+		"/api/consent?consent_challenge=test_challange",
+		nil, handleConsent)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	t.Logf("Result:\n%s\n", string(data))
```
nsklikas, line 273: Same

ory_mocking/Login/login.go
```diff
@@ -0,0 +1,45 @@
...
+
+type LoginUpdateResponse struct {
+	Session       testSession.SessionResponse `json:"session"`
+	Session_token string                      `json:"session_token"`
+}
+
+func SelfServiceLoginHandler(w http.ResponseWriter, r *http.Request) {
+	traits := testSession.IdentityTraits{Name: "TestName"}
+	identity := testSession.SessionIdentity{
+		Id:        "test1234",
```
nsklikas, line 24: Let's not hard code this in the test, other functions may need this value so you would have to hard code the same string to two places which may cause unexpected errors in the future. I think it's best to put it in a global variable, eg:
```go
SESSION_IDENTITY_ID = "test1234"
```

ory_mocking/Login/login.go
```diff
@@ -0,0 +1,45 @@
...
+		SchemaId:  "test_identity.schema.json",
+		SchemaUrl: schemaServerURL,
+		Traits:    traits,
+	}
+	session := testSession.SessionResponse{
+		Id:       "test-1111",
```
nsklikas, line 30: Same

ory_mocking/Consent/consent.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"encoding/json"
+	"log"
+	"net/http"
+)
+
+type OAuth2ConsentResponse struct {
+	Challenge string `json:"challenge"`
+}
```
nsklikas, line 11: Why do we need to define these structs, can't we use the ones in the hydra/kratos sdk?

`hydra_client.OAuth2ConsentRequest` seems to be the same struct. The same is true for the rest of the response structs.

ory_mocking/Testservers/testservers.go
```diff
@@ -0,0 +1,136 @@
...
+
+func createKratosMockServer() *httptest.Server {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/self-service/errors", testErrors.SelfServiceErrorsHandler)
+	mux.HandleFunc("/sessions/whoami", testSession.SessionWhoAmIHandler)
+	mux.HandleFunc("/self-service/login/browser", testLoginBrowser.SelfServiceLoginBrowserHandler)
+	mux.HandleFunc("/self-service/login", testLoginUpdate.SelfServiceLoginHandler)
+
+	s := httptest.NewServer(mux)
+	os.Setenv("KRATOS_PUBLIC_URL", s.URL+"/")
```
nsklikas, line 31:
```suggestion
	os.Setenv("KRATOS_PUBLIC_URL", s.URL)
```

main.go
```diff
@@ -114,7 +114,7 @@ func handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 		session, session_resp, e := kratos.FrontendApi.ToSession(context.Background()).
 			Cookie(cookiesToString(r.Cookies())).
 			Execute()
-
+		//added now
```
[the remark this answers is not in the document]
bencekov: ok

main_test.go
bencekov: I did it for most tests. The error tests are "generated" with a function, which I would rather not pass the t test object to for now.

main_test.go
bencekov: Changed if's to assertions.

main_test.go
bencekov: I rather not. It's true that it's repeated a lot, however:
- If it's ok with you I rather stick to unmarshaling to the type instead of to some map[string]interface{} construct. I think trying with some reflection magic to send in a type for instantiating would make the tests more complicated than they need to be.
- also you don't save much lines at all. Most of these highlighted lines are actually error handling which you would still handle in the tests, because convention (and my preference is) is that functions don't handle errors just send them back to the caller.

main_test.go
bencekov: If it's ok with you, I'd leave the tests as they are. I left comments in main.go, and I'll implement them at some other point.

ory_mocking/Login/login.go
bencekov: I refactored most vars.

ory_mocking/Consent/consent.go
bencekov: I replaced most of the inhouse structs. I kept the ones relating to error messages. The ory libraries store error messages as map[string]interface{} structs, and I already wrote the more comfortable structs. I would use the openapi error constructs in the main.go server.

ory_mocking/Testservers/testservers.go
bencekov: done

ory_mocking/Testservers/testservers.go
```diff
@@ -0,0 +1,131 @@
...
+
+func createSchemaMockServer() *httptest.Server {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/testschema", SchemaHandler)
+	s := httptest.NewServer(mux)
+	schema_server_url = s.URL
+	handlers.SetSchemaServerURL(s.URL)
+	handlers.SetSchemaServerURL(s.URL)
+	return s
+}
```
nsklikas, line 47: Why is this needed?

test_identity.schema.json
```diff
@@ -0,0 +1,17 @@
+{
```
nsklikas, line 1: This file should be placed in the `ory_mocking` folder.

main_test.go
nsklikas: You should pass `t` as a parameter to the test server initialization function, this will allow the server to take care of its own cleanup:
```go 
...
func TestHandleCreateFlowWithoutCookie(t *testing.T) {
        testServers.CreateTestServers(t)
        ...
}
...

func CreateTestServers() func(t *testing.T) {
	kratos = createKratosMockServer(t)
	hydra = createHydraMockServer(t)
}
...

func createKratosMockServer(t *testing.T) *httptest.Server {
       ...
       t.Cleanup(s.Close())
       return s
}
```

ory_mocking/Handlers/handlers.go
```diff
@@ -0,0 +1,219 @@
...
+
+	hydra_client "github.com/ory/hydra-client-go/v2"
+	kratos_client "github.com/ory/kratos-client-go"
+)
+
+const (
+	BROWSER_LOGIN_ID        = "test_id"
+	BROWSER_LOGIN_TYPE      = "browser"
+	OAUTH2_SUBJECT          = "test1234"
+	REDIRECT                = "test.test"
```
nsklikas, line 19: Correct me if I'm wrong, but this is 2 variables `AUTHORIZATION_REDIRECT_URL` and `CONSENT_REDIRECT_URL`.

nsklikas (changes requested):

ory_mocking/Testservers/testservers.go
bencekov: The client sometimes receives an object that includes an object, a schema ID, and a schema URL. The client supposedly (according to documentation) validates the object structure against the schema URL. During debugging when some of the error messages weren't clear I implemented this schema server, and passed it's url into the http responses from the mock server.

main_test.go
bencekov: done. at first I was more in favour of restricting the testing objects to the test_main module, but your solution is valid too. I changed my fork to conform to this.

main_test.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+	responseRedirect := hydra_client.NewOAuth2RedirectToWithDefaults()
+	if err := json.Unmarshal(data, responseRedirect); err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+
+	if responseRedirect.RedirectTo != "test.test" {
+		t.Errorf("expected test.test, got %v", string(data))
+	}
+	assert.Equalf(t, "test.test", responseRedirect.RedirectTo, "Expected %s, got %s.", "test.test", responseRedirect.RedirectTo)
```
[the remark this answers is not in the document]
bencekov: it is. I kinda had to flip flop between refactoring the functions because of cross dependencies in the modules (good call for simplifying module layout btw.) This is a mistake. I fixed it.

ory_mocking/Handlers/handlers.go
```diff
@@ -0,0 +1,219 @@
...
+	TRAITS_NAME             = "TestName"
+	IDENTITY_ID             = "test1234"
+	SCHEMA_ID               = "test_identity.schema.json"
+	SESSION_ID              = "test-1111"
+	ERROR_CODE              = 400
+	ERROR_STATUS            = "Bad Request"
+	ERROR_REASON            = "This is a test"
+	ERROR_MESSAGE           = "This is a test"
+	DEFAULT_ERROR_CODE      = 599
+	TEST_CHALLANGE          = "test_challange"
```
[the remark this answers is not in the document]
bencekov: I agree that test_challange is not very descriptive. I'll change it, and please let me know if you find the other constant names not sufficiently descriptive.

ory_mocking/Handlers/handlers.go
bencekov: fixed

ory_mocking/Testservers/testservers.go
bencekov: I checked and the client doesn't actually validate with a schema received from the url. I'll take the schemaserver out of the testing process, but I'd like to keep the function, and constructs in the code for future reference. According to the doc, the client should verify, that whatever traits object it receives conform to a schema in the schema_id and schema_url parameters.

test_identity.schema.json
bencekov: done

ory_mocking/Testservers/testservers.go
nsklikas: I would think that the schema would be served from Kratos

nsklikas (approved):

.github/workflows/unittest.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+on:
+  workflow_call:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Clean Install of Node version 18
```
gruyaume, line 11: Can't we leverage any of those workflows:
- https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-nodejs
- https://github.com/marketplace/actions/set-up-node-using-nvm

.github/workflows/scan.yaml
```diff
@@ -0,0 +1,20 @@
+name: Scan
```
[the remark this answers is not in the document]
massimilianogo: FYI I shared your [Nginx PR](https://github.com/canonical/nginx-rock/pull/4) with the team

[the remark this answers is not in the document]
ghislainbourgeois: This is exactly what I am proposing in the spec. At some point I expect this workflow to be centralized somewhere, but this is great to start implementing it now and see if we can add improvements!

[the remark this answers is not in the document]
bencekov: sounds good

gruyaume (approved):

.github/workflows/build.yaml
```diff
@@ -7,6 +7,8 @@ on:
       - main
```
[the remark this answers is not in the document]
bencekov: done

PR #23, opened 2023-04-24
Fix rockcraft.yaml

nsklikas:
Fix `rockcraft.yaml`:
- Fix license.
- Replace `entrypoint` with `services`

### DIFF
--- a/rockcraft.yaml
+++ b/rockcraft.yaml
@@ -6,7 +6,7 @@ summary: Canonical Identity platform login UI
 description: |
   This is the Canonical Identity platform login UI used for connecting
   Ory Kratos with Ory Hydra.
-license: GPL-3.0
+license: Apache-2.0
 platforms:
   amd64:
 
--- a/rockcraft.yaml
+++ b/rockcraft.yaml
@@ -9,7 +9,12 @@ description: |
 license: GPL-3.0
 platforms:
   amd64:
-entrypoint: ["identity_platform_login_ui"]
+
+services:
+  login-ui:
+    override: replace
+    command: identity_platform_login_ui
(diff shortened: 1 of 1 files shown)


gruyaume (approved):

rockcraft.yaml
```diff
@@ -6,10 +6,15 @@ summary: Canonical Identity platform login UI
...
+license: GPL-3.0
 platforms:
   amd64:
-entrypoint: ["identity_platform_login_ui"]
+
+services:
+  login-ui:
+    override: replace
+    command: identity_platform_login_ui
+    startup: enabled
```
natalian98, line 17: Why we need to replace entrypoint with services? Will it not interfere with `services` in [pebble layer](https://github.com/canonical/identity-platform-login-ui-operator/blob/main/src/charm.py#L129)?

rockcraft.yaml
```diff
@@ -9,7 +9,12 @@ description: |
 license: GPL-3.0
```
natalian98, line 9: could there be a problem with your branch? I see the license is changed in both commits, but overall in PR's "Files changed" it's not modified

rockcraft.yaml
massigori: @natalian98 that seems correct as per the layer specification https://github.com/canonical/pebble#layer-specification

rockcraft.yaml
nsklikas: Ok this is very weird, I force pushed my commits again and it seems fixed now.

natalian98 (dismissed):

bencekov (approved):
Looks good.

PR #27, opened 2023-05-12
Use relative paths in redirects

nsklikas:
When the UI is served behind a reverse proxy (e.g. traefik), the pages may be served under a subpath. This breaks our redirects and calls to the backend API. In order to overcome this we use relative paths everywhere.

This is not the proper solution and ideally we would want the reverse proxy to inject the html with the base URL tag.

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -109,7 +109,7 @@ const Login: NextPage = () => {
               <Card title="Choose Provider" >
                 <Flow onSubmit={onSubmit} flow={flow} />
               </Card> :
-              <Spinner></Spinner>
+              <Spinner />
             }
           </div>
         </div>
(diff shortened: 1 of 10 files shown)


ui/pages/login.tsx
```diff
@@ -102,12 +105,12 @@ const Login: NextPage = () => {
...
-          <Card title="Choose Provider" >
-            <Flow onSubmit={onSubmit} flow={flow} />
-          </Card>:
-          <Spinner></Spinner>
-          }
+            {flow ?
+              <Card title="Choose Provider" >
+                <Flow onSubmit={onSubmit} flow={flow} />
+              </Card> :
+              <Spinner></Spinner>
```
lorumic, line 112: We don't have any content inside this component, so we can use a self-closing tag here:
```suggestion
              <Spinner />
```

lorumic (commented):
I cannot help with the traefik part as I don't really know much about it, but maybe @edlerd has an opinion about this relative path issue?

Apart from that, there is only one small "code style" suggestion that I'm adding below.

edlerd (dismissed):
LGTM
Have upstream provided a `BASE_URL` via config (or environment) might be safer for the future.

natalian98 (dismissed):

bencekov (approved):
Looks good!

PR #30, opened 2023-05-30
Iam 307 implement health checks

bencekov:
This PR implements liveness and readiness checks for the Login UI server. Readiness response is only available if dependencies are also available.

Liveness check is available at /health/alive. This check returns with {"status": "ok"} if the server is running. Currently the endpoint isn't configured to return with other values, so a http timeout error during liveness check implies the server is down.

Readiness check is available at /health/ready. This check returns with {"status": "ok"} if the server is running. Currently the endpoint isn't configured to return with other values, so a http timeout error during readiness check implies the server is down.

EDIT: Updated the logic in responding to /health/ready
EDIT 2: Updated the logic in responding to /health/ready so ready=ok means the server is running

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -1,39 +1,27 @@
 package health
 
 import (
-	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
-
-	hydra_client "github.com/ory/hydra-client-go/v2"
-	kratos_client "github.com/ory/kratos-client-go"
 )
 
 const okValue = "ok"
+const kratosEnvar = "KRATOS_PUBLIC_URL"
+const hydraEnvar = "HYDRA_ADMIN_URL"
 
 var aliveSingleton Status
 var readySingleton Status
-var kratos *kratos_client.APIClient
-var hydra *hydra_client.APIClient
-var apiClientSet bool = false
 
 type Status struct {
 	Status  string `json:"status"`
 	Message string `json:"message,omitempty"`
 }
 
-func SetApiClients(k *kratos_client.APIClient, h *hydra_client.APIClient) {
-	kratos = k
-	hydra = h
-	apiClientSet = true
-}
-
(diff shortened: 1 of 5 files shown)


health/health.go
```diff
@@ -0,0 +1,165 @@
...
+
+func TestHandleAlive(w http.ResponseWriter, r *http.Request) {
+	handleAlive(w, r)
+}
+
+func TestHandleReady(w http.ResponseWriter, r *http.Request) {
+	handleReady(w, r)
+}
+
+func TestResetHealth() {
```
gruyaume, line 159: Why are those "Test" functions directly in `health.go`?

gruyaume:
Can you please explain the logic implemented here (or point to a spec)? From what I understand:

The login UI will serve two new endpoints:
- `/health/ready`: Returns whether both hydra and kratos return true when making API calls to their "IsReady" endpoints
- `/health/alive`: It's still unclear to me what could set the "alive" state to False looking at the code.

natalian98:
> Can you please explain the logic implemented here (or point to a spec)? From what I understand:
> 
> The login UI will serve two new endpoints:
> 
> * `/health/ready`: Returns whether both hydra and kratos return true when making API calls to their "IsReady" endpoints

@bencekov if that's the case, why do we care if hydra and kratos work in the context of login ui? They already both have this health check independently. Shouldn't we rather check if login ui itself is up and running?

health/health.go
bencekov: The states for health checks are encapsulated in the package. I tried to keep access to them restrictive so that the health package is more like an API, and not a library. However this makes writing unit tests inconvenient. The functions starting with Test are either wrappers around private methods, or they access private variables. The Test prefix implies that these are public methods only intended to use for testing.

bencekov:
The PR https://github.com/canonical/identity-platform-login-ui/pull/33 replaces this PR.

PR #32, opened 2023-06-01
IAM-308-Implement Prometheus Wrapper

bencekov:
This PR implements prometheus instrumentation for the login UI server. The Metrics struct in metrics.go, and the MiddlewareManager in middleware.go are heavily inspired by ory/x/prometheusx solution for implementing the Collector interface in the prometheus go client, and a middleware that updates counters based on the requested handlers.

The handler for the metrics is available in the handler.go as the function PrometheusMetrics.
The instrumentation happens via a middleware, which calls on the Instrument method of the prometheusMetrics property. The endpoint label for the metrics has to be supplied manually, hence the need for registering handler paths, and the function getLabelForPath. I verified that of the Ory APIs that we proxy for, none uses path parameters, which would have to be handled in the getLabelForPath function. (Having path variables would increase the cardinality of labels)

More detail [here](https://docs.google.com/document/d/1bj11M5lepbYnP19PE3OQScbcXZM3xgz8Kq44rUQI88I/edit#heading=h.4pclarsut2af)

Edit:
Added unit tests in prometheus folder.
Added another middleware for extracting status code from ResponseWriter.


gruyaume:
Please add a description to the PR

prometheus/metrics.go
```diff
@@ -0,0 +1,152 @@
...
+		switch {
+		case status >= 200 && status <= 299:
+			statusBucket = "2xx"
+		case status >= 300 && status <= 399:
+			statusBucket = "3xx"
+		case status >= 400 && status <= 499:
+			statusBucket = "4xx"
+		case status >= 500 && status <= 599:
+			statusBucket = "5xx"
+		}
```
nsklikas, line 119: I am not sure if we want to do this. Would it create such a high cardinality if we just used the status as a label?

For example I would not want to get `400-404` mixed up with `422` (which is what we get every time the user chooses their provider.

main_test.go
```diff
@@ -313,3 +319,74 @@ func CreateGenericTest(t *testing.T, serverCreater func(t *testing.T), HttpMetho
 	}
 	return data, nil
 }
+
+// --------------------------------------------
+// TESTING 	PROMETHEUS INSTRUMENTATION
+// --------------------------------------------
+func TestHandlePrometheusInstrumentation(t *testing.T) {
```
nsklikas, line 326: Shouldn't this test be moved in the `prometheus` dir?

main.go
```diff
@@ -342,3 +347,24 @@ func getUserClaims(i kratos_client.Identity, cr hydra_client.OAuth2ConsentReques
...
+		"/api/kratos/self-service/login/flows",
+		"/api/kratos/self-service/login",
+		"/api/kratos/self-service/errors",
+		"/api/consent",
+		"/consent.html",
+		"/error.html",
+		"/index.html",
+		"/login.html",
+		"/",
+		"",
```
nsklikas, line 364: What is the difference between these 2?

main.go
```diff
@@ -342,3 +347,24 @@ func getUserClaims(i kratos_client.Identity, cr hydra_client.OAuth2ConsentReques
...
+		"/oidc_error",
+		"/registration",
```
nsklikas, line 366: I don't really like having to manually add each path to this list, it seems error prone.

For the server paths, couldn't we use the middleware to register them?
For the UI paths it is not that simple (we could do it on runtime, but that is not ideal). Perhaps we could scan the UI dir on startup and construct this list?

nsklikas (commented):
Good job, looks good overall. I left some comments but I have not run it yet, once I get to that I will leave some more (probably).

I understand some of the code is taken from the `oryx` repo, could you add a comment to the relevant parts of the code with a link to the upstream file?

main_test.go
bencekov: If you want, I can add separate unit tests for the objects in the prometheus dir, but this test calls the handlers defined in main, so it should stay in main_test.go.

main.go
bencekov: I thought the endpoint was gonna show up differently, but verified that it shows up as the same

bencekov:
I'll put this in the document I'm writing currently, but here's a heads up:
 
The original implementation of this path thing includes a filter that filters out path parameters and in the metrics replaces them with {parameter} to reduce metric cardinality. Ory uses httprouter instead of net/http which has some features that help dealing with path parameters. More importantly, before adding new proxies to ory API that use path parameters, we will have to implement functionality to extract said path parameters. I verified that the current API we use don't make use of path parameters.

main.go
bencekov: For your second point. Sorry but what UI dir? I was under the impression that we embed everything into a single binary when deploying.

main.go
```diff
@@ -342,3 +347,37 @@ func getUserClaims(i kratos_client.Identity, cr hydra_client.OAuth2ConsentReques
...
+	mm.RegisterRoutes(
+		"/api/kratos/self-service/login/browser",
+		"/api/kratos/self-service/login/flows",
+		"/api/kratos/self-service/login",
+		"/api/kratos/self-service/errors",
+		"/api/consent",
+		prometheus.PrometheusPath,
+	)
+
+	pages, err := ui.ReadDir("ui/dist/_next/static/chunks/pages")
```
nsklikas, line 362: This is not correct, only `js` files are in that folder. You need to register the all the files with the `.html` suffix in `ui/dist`. 

I thought we should be able to reuse the object from https://github.com/canonical/identity-platform-login-ui/pull/32/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R84 to do this

bencekov: Sorry, I somehow didn't notice the html pages in dist. I fixed that. So I have 2 reservations against using the middleware to "register" these endpoints:
- I want the package prometheus to do one thing well, which in this case is having a middleware to implement the instrumentation of http handlers. Having the middleware do scan the dir directory would adds complexity to an otherwise straightforward package. (I try to make packages as if they were separate modules with API to access their functionality. I think it makes the code base more manageable in the long run.)
- I think a middleware is well suited for work you need to do in the runtime every time you call the decorated function. In this case, scanning the ui/dist dir is something you only need to do at startup, because in our use case none of our pages should change in runtime.

Also I updated MetricsManager to store routes in a map instead of slice.

prometheus/middleware.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"net/http"
+)
+
+type MetricsManager struct {
+	prometheusMetrics *Metrics
+	routes            map[string]bool
+}
+
+func NewMetricsManager(app, version, hash, buildTime string) *MetricsManager {
```
gruyaume, line 12: Where is this used?

prometheus/middleware.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+)
+
+type MetricsManager struct {
+	prometheusMetrics *Metrics
+	routes            map[string]bool
+}
+
+// NewMetricsManagerWithPrefix creates MetricsManager that uses metricsPrefix parameters as a prefix
+// for all metrics registered within this middleware. Setting empty string in metricsPrefix will be equivalent to calling NewMetricsManager.
```
gruyaume, line 13: There is no `NewMetricsManager` method

prometheus/handler_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	req := httptest.NewRequest(http.MethodGet, handler_testPath, nil)
+	req.Header.Set("Content-Type", "application/json")
+	w := httptest.NewRecorder()
+	mm.Middleware(PrometheusMetrics)(w, req)
+	resp := w.Result()
+	assert.Equalf(t, http.StatusOK, resp.StatusCode, "Expected %s, got %s", http.StatusOK, resp.StatusCode)
+	textParser := expfmt.TextParser{}
+	text, err := textParser.TextToMetricFamilies(resp.Body)
+	assert.Nilf(t, err, "Expected nil, got %s", err)
```
gruyaume, line 28: Are those asserts all over this method standard practice?

main.go
```diff
@@ -85,13 +90,15 @@ func main() {
...
-	http.HandleFunc("/api/kratos/self-service/login", handleUpdateFlow)
-	http.HandleFunc("/api/kratos/self-service/errors", handleKratosError)
-	http.HandleFunc("/api/consent", handleConsent)
+	//status_code.ResponseWithStatusCodeMiddleware()
+	http.HandleFunc("/api/kratos/self-service/login/browser", metricsManager.Middleware(handleCreateFlow))
+	http.HandleFunc("/api/kratos/self-service/login/flows", metricsManager.Middleware(handleLoginFlow))
+	http.HandleFunc("/api/kratos/self-service/login", metricsManager.Middleware(handleUpdateFlow))
+	http.HandleFunc("/api/kratos/self-service/errors", metricsManager.Middleware(handleKratosError))
+	http.HandleFunc("/api/consent", metricsManager.Middleware(handleConsent))
+	http.HandleFunc(prometheus.PrometheusPath, metricsManager.Middleware(prometheus.PrometheusMetrics))
```
shipperizer, line 101: so the fact that we are using the DefualtServeMux means that we are missing out on loads of features coming from "router" libs, like the `Use` method on chi, gorilla/mux etc

Might be useful having a chat before merging this

main_test.go
bencekov: unit tests added

shipperizer:
@bencekov sovled conflicts and moved to https://github.com/canonical/identity-platform-login-ui/pull/43

please take a look at that if happy and we'll get it in

shipperizer (approved):

PR #33, opened 2023-06-08
IAM-307 Liveness check implementation

bencekov:
The purpose of this PR is to implement a liveness check for Login UI. It is accessible on the endpoint /health/alive. If the server is running, liveness check returns with '{'status':'ok'}'.

The pr includes the health.go library that exposes the handler for the liveness check. It also includes a handler for a readiness check which is not used in main, and is only there for possible future usage.

The pr also includes unit tests for the liveness check.

Additionally the pr includes mocked health checks for the mocked ory servers used for unit tests.

- health library implementing an API to store alive and ready statuses independently
- main.go implementing liveness check only
- tests liveness checks
- Handlers for mocking health checks
- mocked servers implementing health check mocking

This pr replaces https://github.com/canonical/identity-platform-login-ui/pull/30

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -1,8 +1,5 @@
 package health
 
-//Notes on Thread safety: Currently we only expose methods that read the state of the singletons therefore thread safety is not an issue.
-//In case the scope of use case for Health expands: Implement RWLocks for the singleton variables, or implement using buffered channels to change state
-
 import (
 	"encoding/json"
 	"log"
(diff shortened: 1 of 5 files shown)


health/health.go
```diff
@@ -0,0 +1,97 @@
...
+
+func SetAlive() {
+	aliveSingleton = Status{Status: okValue}
+}
+
+func HandleAlive(w http.ResponseWriter, r *http.Request) {
+	status := getAlive()
+	w.Header().Set("Content-Type", "application/json")
+	if status.Status == okValue {
+		w.WriteHeader(200)
```
shipperizer, line 62: small thing, not a biggie, but the net.http package offers these values as const 

https://go.dev/src/net/http/status.go

main_test.go
```diff
@@ -313,3 +315,46 @@ func CreateGenericTest(t *testing.T, serverCreater func(t *testing.T), HttpMetho
...
+	req := httptest.NewRequest(http.MethodGet, HANDLE_ALIVE_URL, nil)
+	w := httptest.NewRecorder()
+	health.HandleAlive(w, req)
+	res := w.Result()
+	defer res.Body.Close()
+	data, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		t.Errorf("expected error to be nil got %v", err)
+	}
+	recievedStatus := health.EmptyStatus()
```
shipperizer, line 333:
```suggestion
	receivedStatus := health.EmptyStatus()
```

shipperizer (commented):
@bencekov just to open a discussion, what's our policy with mocks?do we commit them or we auto-generate them on the fly?

Last point, commit messages, I see.you follow the conventional commit which is great, if the message is a bit more specific (not detailed) like `feat: setup healthcheck handler` that helps a great deal when tracking down features and bugs on the changeling

All in all code is clean, so good job

bencekov:
We commit the mocks. For the operators we use the Magicmock, and for this repo I wrote the mock handlers and use them with net/httptest mock servers.

health/health.go
```diff
@@ -0,0 +1,97 @@
...
+	} else {
+		w.WriteHeader(503)
+	}
+	body, err := json.Marshal(status)
+	if err != nil {
+		log.Printf("Error during Health Check Readiness Check\nerror: %s", err.Error())
+	}
+	w.Write(body)
+	return
```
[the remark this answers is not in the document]
bencekov: I'll just simplify the health package, and make it stateless.

health/health.go
```diff
@@ -0,0 +1,51 @@
...
+	return Status{Status: okValue}
+}
+
+func HandleAlive(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	body, err := json.Marshal(GetOKStatus())
+	if err != nil {
+		log.Printf("Error during Health Check Liveness Check\nerror: %s", err.Error())
+	}
```
shipperizer, line 33: technically the error is not hampering the healthcheck, so the log is misleading

unless u cahnge response when the `json.Marshal` fails

health/health.go
```diff
@@ -0,0 +1,51 @@
...
+	Status  string `json:"status"`
+	Message string `json:"message,omitempty"`
+}
+
+func GetNotOKStatus(msg string) Status {
+	return Status{
+		Status:  http.StatusText(http.StatusServiceUnavailable),
+		Message: msg,
+	}
+}
```
shipperizer, line 21: this is not being used, i'd say just drop it

health/health.go
```diff
@@ -0,0 +1,51 @@
...
+		log.Printf("Error during Health Check Liveness Check\nerror: %s", err.Error())
+	}
+	w.Write(body)
+	return
+}
+
+func HandleReady(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+	body, err := json.Marshal(GetOKStatus())
```
shipperizer, line 41: i'd say keep it simple
```suggestion
	body, _ := json.Marshal(Status{Status:fmt.Sprintf(http.StatusOK), Message: okvalue})
```

wouldn't even bother about checking the error in this case

shipperizer (approved):
:shipit:

PR #42, opened 2023-06-16
Logging Package

bencekov:
Logging Package and Unit tests


shipperizer (commented):
@bencekov did you follow any specific spec or example for this?

if not, let's have a chat around this, there are libraries around structured logging, for example https://github.com/uber-go/zap or https://github.com/rs/zerolog

we should pivot towards that

shipperizer (approved):

PR #43, opened 2023-06-19
IAM 308/prometheus

shipperizer:
- feat: Add Ory based wrapper object to LoginUI
- feat: implemented login ui prometheus instrumentation without unit tests
- feat: Added unit test for Prometheus instrumentation + fixed minor url bug
- fix:removed "" from registered paths for metrics instrumentation labels
- feat: app scans ui pages dir to register to prometheus instrumentation
- fix: fixed registerHelper function in main
- feat: routes in MetricsManager in package prometheus are in map instead of list
- feat: added alive check to routes registered with prometheus middleware
- feat: added internal middleware status_code to prometheus package to provide observability to returned status codes and related unit tests
- fix: fix comment, change visibility of Cleanup function in prometheus package
- feat: metadata extraction from http response functions are now in package http_meta


bencekov (approved):
- Verified that unit tests run.
- My only comment is that in main the handlers don't seem to use the http_meta middleware

Looks good! Good luck with the rewrite/refactor!

PR #46, opened 2023-06-19
IAM-314-Add piping, shell, and tee to rock

bencekov:
Added components to rock, that are necessary to logging with our current methods.

### DIFF
--- a/rockcraft.yaml
+++ b/rockcraft.yaml
@@ -17,6 +17,20 @@ services:
     startup: enabled
 
 parts:
+  util:
+    plugin: nil
+    stage-packages:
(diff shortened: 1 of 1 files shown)


nsklikas (dismissed):

shipperizer:
@bencekov had a chat with the COS guys, so was wondering if we could simply pipe our log outputs to stdout and "a file" and then point promtail to it instead of getting sh and tee?

this is good to merge anyway i guess, we can always pull it out if not needed

nsklikas:
This was implemented in the application code after all (https://github.com/canonical/identity-platform-login-ui/commit/dc763681a85c492121e163e1526b901a75ca3849)

PR #73, opened 2023-06-30
Separate logic in pkg/kratos

nsklikas:
IAM-326: split http layer from business logic
 
- Separate api/logic in `pkg/kratos`
- Add tests


internal/misc/http/helpers.go
```diff
@@ -21,41 +22,41 @@ func GetBaseURL(r *http.Request) string {
...
+	}
+	return nil
+}
+
+func WriteHeaders(w http.ResponseWriter, headers http.Header) {
+	for k, vs := range headers {
 		for _, v := range vs {
 			w.Header().Set(k, v)
 		}
 	}
```
shipperizer, line 45: i think that even if it get's duplicated in different packages, this `func WriteHeaders` can become a private method in any `API` and we keep it in the handlers file

internal/misc/http/helpers.go
```diff
@@ -21,41 +22,41 @@ func GetBaseURL(r *http.Request) string {
...
+	var ret = make([]string, len(cookies))
 	for i, c := range cookies {
 		ret[i] = fmt.Sprintf("%s=%s", c.Name, c.Value)
 	}
 	return strings.Join(ret, "; ")
 }
 
-func ParseBody(r *http.Request, body interface{}) *interface{} {
-	decoder := json.NewDecoder(r.Body)
+func ParseBody(b io.ReadCloser, body interface{}) error {
```
shipperizer, line 56: same as for `WriteHeader`, better be explicit and have simple redundant code (one liners in most cases) than a package that delivers really small value

doesn't need to be part of this PR, but simply add a `TODO` and open an issue

pkg/kratos/interfaces.go
```diff
@@ -1,14 +1,28 @@
...
+}
+
+type ServiceInterface interface {
+	CheckSession(context.Context, []*http.Cookie) (*kClient.Session, http.Header, error)
+	AcceptLoginRequest(ctx context.Context, identityID string, lc string) (*hClient.OAuth2RedirectTo, http.Header, error)
+	CreateBrowserLoginFlow(ctx context.Context, aal string, return_to string, login_challenge string, refresh bool, cookies []*http.Cookie) (*kClient.LoginFlow, http.Header, error)
+	GetLoginFlow(ctx context.Context, id string, cookies []*http.Cookie) (*kClient.LoginFlow, http.Header, error)
+	UpdateOIDCLoginFlow(ctx context.Context, flow string, body kClient.UpdateLoginFlowBody, cookies []*http.Cookie) (*ErrorBrowserLocationChangeRequired, http.Header, error)
+	GetFlowError(ctx context.Context, id string) (*kClient.FlowError, http.Header, error)
+	parseLoginFlowMethodBody(r *http.Request) (*kratos_client.UpdateLoginFlowBody, error)
```
shipperizer, line 27:
```suggestion
	ParseLoginFlowMethodBody(r *http.Request) (*kratos_client.UpdateLoginFlowBody, error)
```

shipperizer, line 27:
```suggestion
	AcceptLoginRequest(context.Context, string, string) (*hClient.OAuth2RedirectTo, http.Header, error)
	CreateBrowserLoginFlow(context.Context, string, string, string, bool, []*http.Cookie) (*kClient.LoginFlow, http.Header, error)
	GetLoginFlow(context.Context, string, []*http.Cookie) (*kClient.LoginFlow, http.Header, error)
	UpdateOIDCLoginFlow(context.Context, string, kClient.UpdateLoginFlowBody, []*http.Cookie) (*ErrorBrowserLocationChangeRequired, http.Header, error)
	GetFlowError(context.Context, string) (*kClient.FlowError, http.Header, error)
	parseLoginFlowMethodBody(*http.Request) (*kratos_client.UpdateLoginFlowBody, error)
```

convention goes that u dont name args in interfaces (works anyway in fairness)

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+	kratos KratosClientInterface
+	hydra  HydraClientInterface
+
+	tracer  trace.Tracer
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+// We override the type from the kratos sdk, as it does not
+// get marshalled correctly into json
```
shipperizer, line 25: what is the problem with https://pkg.go.dev/github.com/ory/kratos-client-go#GenericError ?

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (s *Service) AcceptLoginRequest(ctx context.Context, identityID string, lc string) (*hClient.OAuth2RedirectTo, http.Header, error) {
+	_, span := s.tracer.Start(ctx, "hydra.OAuth2Api.AcceptOAuth2LoginRequest")
+	defer span.End()
+
+	accept := hClient.NewAcceptOAuth2LoginRequest(identityID)
+	redirect_to, resp, err := s.hydra.OAuth2Api().AcceptOAuth2LoginRequest(ctx).
+		LoginChallenge(lc).
+		AcceptOAuth2LoginRequest(*accept).
+		Execute()
```
shipperizer, line 59: run `gofmt -w .` it will sort out the indentation/style

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if err != nil {
+		// TODO @shipperizer we shouldn't be logging this
```
shipperizer, line 62:
```suggestion
		// TODO @nsklikas  we shouldn't be logging this
```

:dancers:

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+		s.logger.Debugf("full HTTP response: %v", resp)
+		return nil, nil, err
+	}
+
+	return redirect_to, resp.Header, nil
+}
+
+func (s *Service) CreateBrowserLoginFlow(
+	ctx context.Context, aal string, return_to string, login_challenge string, refresh bool, cookies []*http.Cookie,
```
shipperizer, line 71:
```suggestion
	ctx context.Context, aal, return_to, login_challenge string, refresh bool, cookies []*http.Cookie,
```

sugar syntax or shortcut (still kinda normal practice afaik), if subsequent args are the same type, u can specify only on the last one the type

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+	_, span := s.tracer.Start(ctx, "kratos.FrontendApi.CreateBrowserLoginFlow")
+	defer span.End()
+
+	flow, resp, err := s.kratos.FrontendApi().
+		CreateBrowserLoginFlow(context.Background()).
+		Aal(aal).
+		ReturnTo(return_to).
+		LoginChallenge(login_challenge).
+		Refresh(refresh).
+		Cookie(misc.CookiesToString(cookies)).
```
shipperizer, line 82: 10£ if we can get rid of the `misc` package, same reason as https://github.com/canonical/identity-platform-login-ui/pull/73/files#r1248104788

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		// TODO @shipperizer we shouldn't be logging this
+		s.logger.Debugf("full HTTP response: %v", resp)
+
+		return nil, nil, err
+	}
+	return session, resp.Header, nil
+}
+
+func (s *Service) AcceptLoginRequest(ctx context.Context, identityID string, lc string) (*hClient.OAuth2RedirectTo, http.Header, error) {
```
shipperizer, line 51: i see what you meant when saying that the return of the headers is a bit ugly, other way would be to apss in a `Writer` and change it inside the function....I mean not a bad thing but you'd incure in the same pain that we are having right now with testing these type of funcs https://github.com/canonical/identity-platform-login-ui/pull/73/files#diff-5e4ba31969008b305d67765d5d4c9bd50d9bcc49571f1594e97d2b152ddf1378R38-R39

all in all for now we are ok with 3 returns, worst case we can return a struct composed by the library object and the headers

```
type X struct {
    Headers http.Header
    Object interface{}
}
``` 

just to make it generic, but it looks terrible, which on a scale it's probably worse than ugly

pkg/kratos/service.go
```diff
@@ -0,0 +1,171 @@
...
+		Flow(flow).
+		UpdateLoginFlowBody(body).
+		Cookie(misc.CookiesToString(cookies)).
+		Execute()
+	if err != nil && resp.StatusCode != 422 {
+		s.logger.Debugf("full HTTP response: %v", resp)
+		return nil, nil, err
+	}
+
+	redirect_resp := new(ErrorBrowserLocationChangeRequired)
```
shipperizer, line 126: `camelCase` instead of `snake_case`

pkg/kratos/service_test.go
```diff
@@ -0,0 +1,635 @@
...
+
+	"github.com/canonical/identity-platform-login-ui/internal/monitoring"
+	gomock "github.com/golang/mock/gomock"
+	hClient "github.com/ory/hydra-client-go/v2"
+	kClient "github.com/ory/kratos-client-go"
+	"go.opentelemetry.io/otel/trace"
+)
+
+//go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_logger.go -source=../../internal/logging/interfaces.go
+//go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_kratos_pkg.go -source=./interfaces.go
```
shipperizer, line 22:
```suggestion
//go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_interfaces.go -source=./interfaces.go
```

simply to avoid having the package name in here, I should have done it in here https://github.com/canonical/identity-platform-login-ui/blob/main/pkg/extra/handlers_test.go#L19 too

shipperizer (commented):
really good, let's see the comments together next week

commits wise, the more you split the better,  to review, to see the thoughts flow and plus all the random chit chat about isolation and whatever, happy to have a chat around that but not a blocker

pkg/kratos/service.go
nsklikas: Maybe I am missing something. The struct in the sdk (https://pkg.go.dev/github.com/ory/kratos-client-go#ErrorGeneric) follows the convention described in the docs (https://www.ory.sh/docs/kratos/reference/api#tag/frontend/operation/updateLoginFlow), which is that the error should be like:
```json
{
  "error": {
    "error": {
      "code": 404,
      "debug": "SQL field \"foo\" is not a bool.",
      "details": {},
      "id": "string",
      "message": "The resource could not be found",
      "reason": "User with ID 1234 does not exist.",
      "request": "d7ef54b1-ec15-46e6-bccb-524b82c035e6",
      "status": "Not Found"
    }
  },
  "redirect_browser_to": "string"
}
```

BUT the response I get when running Kratos in docker(`v0.13.0`) locally, the response I get is:
```json
{
  "error": {
    "id": "browser_location_change_required",
    "code": 422,
    "status": "Unprocessable Entity",
    "reason": "In order to complete this flow please redirect the browser to: somewhere",
    "message": "browser location change required"
  },
  "redirect_browser_to": "somewhere"
}
```
The second one is missing one nesting level. I think that this happens because https://github.com/ory/kratos/blob/08fed36973274ef294491d00811bc867f1537d62/selfservice/flow/error.go#L167 is different from the one in the sdk. 

We could simply return a `RedirectBrowserTo` (without the `error` part) if we wanted, so that we don't have to define our own structure.

cmd/main.go
```diff
@@ -52,7 +52,7 @@ func main() {
 	kClient := ik.NewClient(specs.KratosPublicURL)
 	hClient := ih.NewClient(specs.HydraAdminURL)
 
-	router := web.NewRouter(kClient, hClient, distFS, tracer, monitor, logger)
+	router := web.NewRouter(kClient, hClient, distFS, tracer, monitor, logger, specs.BaseURL)
```
shipperizer, line 55:
```suggestion
	router := web.NewRouter(specs.BaseURL, kClient, hClient, distFS, tracer, monitor, logger)
```

as a rule of thumb, let s try to keep the "auxiliary" parts at the end, aka, `tracer`, `monitor` and `logger`

pkg/kratos/service.go
shipperizer: great let's just add a permalink to this comment or add extra on the comment

shipperizer (approved):

PR #74, opened 2023-06-30
IAM-318

shipperizer:
IAM-318:  add log folder and log file to the rock


nsklikas:
I used the `base-files_var` chisel to create the `/var/log` dir (https://github.com/canonical/chisel-releases/blob/ubuntu-22.04/slices/base-files.yaml). We can store our logs there. To test it I exported to image to docker, run it and then used `docker cp` to get the log file and inspect its contents.

nsklikas (approved):

PR #97, opened 2023-07-18
feat: sync version in rockcraft.yaml when release

wood-push-melon:
We would like to publish the ROCK image with the tagged version whenever there is a new release. But this had been accomplished. This pull request will further improve the release flow which will sync the tagged version to the `rockcraft.yaml` file. It also includes some cleanups in the GitHub actions.


rockcraft.yaml
```diff
@@ -2,7 +2,7 @@ name: identity-platform-login-ui
 
 base: bare
 build-base: ubuntu:22.04
-version: '0.1'
+version: '0.6.1' # x-release-please-version
```
wood-push-melon, line 5: This is required by the `google-release-please`. Please refer to this [doc](https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files).

wood-push-melon: Also, we need to update the `version` in the `rockcraft.yaml` file to be the most updated version. Otherwise, the `google-release-please` action won't do any update. The `version` field will be automatically updated in the release pull request.

.github/workflows/publish.yaml
```diff
@@ -30,21 +30,25 @@ jobs:
...
+          versions=(stable "${{ github.ref_name }}")
+        fi
+
+        for version in "${versions[@]}"; do
+          sudo skopeo --insecure-copy \
+            --dest-creds "${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}" \
+            oci-archive:$(realpath ./"${{ inputs.rock }}") \
+            docker://ghcr.io/canonical/identity-platform-login-ui:"${version}"
+          echo "image=ghcr.io/canonical/identity-platform-login-ui:${version}" >> "$GITHUB_ENV"
+        done
```
nsklikas, line 50: Just to be sure: this does not alter the functionality of the upload step, it just merges the 2.

Do you think that having the `if` in shell is clearer than using `if` of the workflow syntax? I think that the `if` in the workflow provides more readability (different name depending on the channel), but I don't have a strong preference either way (cc @shipperizer )

nsklikas (commented):
Good job, LGTM. Only a minor comment about the workflow's `job.step`, which you can disregard

.github/workflows/publish.yaml
shipperizer: i'd prefer the if in the workflow as well tbh, even if it causes repetition

shipperizer (approved):

PR #105, opened 2023-07-28
IAM-386-Fix-Route-Matching bug

bencekov:
Fixed bug, where the server treats all misc. routes as fileserver requests.

Try it out by compiling the app, and requesting different routes.

After that if you access /api/v0/metrics, you'll see that the server only added the .html extension to the routes corresponding to the embedded html files.

### DIFF
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -1,30 +1,36 @@
 package ui
 
 import (
+	"fmt"
 	"io/fs"
 	"net/http"
 	"path"
+	"strings"
 
 	"github.com/canonical/identity-platform-login-ui/internal/logging"
 	"github.com/go-chi/chi/v5"
 )
 
+const UI = "/ui"
+
 type API struct {
 	fileServer http.Handler
-
-	logger logging.LoggerInterface
+	logger     logging.LoggerInterface
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	// TODO @shipperizer unsure if we deal with any POST/PUT/PATCH via js
-	mux.HandleFunc("/*", a.uiFiles)
+	mux.HandleFunc(fmt.Sprintf("%s/*", UI), a.uiFiles)
+
 }
 
 // TODO: Validate response when server error handling is implemented
 func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
-	if ext := path.Ext(r.URL.Path); ext == "" && r.URL.Path != "/" {
-		r.URL.Path += ".html"
+	if ext := path.Ext(r.URL.Path); ext == "" {
+		r.URL.Path = fmt.Sprintf("%s.html", r.URL.Path)
 	}
+
+	r.URL.Path = strings.TrimPrefix(r.URL.Path, UI)
 
 	a.fileServer.ServeHTTP(w, r)
 }
--- a/ui/api/hydra.ts
+++ b/ui/api/hydra.ts
@@ -3,7 +3,7 @@ import { Configuration, OAuth2Api } from "@ory/client";
 const hydraAdmin = new OAuth2Api(
   new Configuration({
     // Use relative path so that this works when served in a subpath
-    basePath: "./api/hydra",
+    basePath: "../api/hydra",
     baseOptions: {
       withCredentials: true,
     },
--- a/ui/api/kratos.ts
+++ b/ui/api/kratos.ts
@@ -3,7 +3,7 @@ import { Configuration, FrontendApi } from "@ory/client";
 export const kratos = new FrontendApi(
   new Configuration({
     // Use relative path so that this works when served in a subpath
-    basePath: "./api/kratos",
+    basePath: "../api/kratos",
     baseOptions: {
       withCredentials: true,
     },


pkg/ui/handlers.go
```diff
@@ -1,23 +1,32 @@
...
-	logger logging.LoggerInterface
+	fileSystem fs.FS
+	logger     logging.LoggerInterface
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	// TODO @shipperizer unsure if we deal with any POST/PUT/PATCH via js
-	mux.HandleFunc("/*", a.uiFiles)
+	nodeList, _ := fs.ReadDir(a.fileSystem, ".")
+	r, _ := regexp.Compile(`html$`)
```
nsklikas, line 24: This does not work, we need to serve other types of files as well like `.js`, `.css` and images. If you go to the `login` page you will see that it does not load.

nsklikas (changes requested):

pkg/ui/handlers.go
shipperizer, line 24: if u got to compile it and reuse it (even though in practice we call it only once, but potentially we could multiple times), better to put it in the `constructor NewAPI`

pkg/ui/handlers.go
```diff
@@ -1,23 +1,32 @@
...
+	for _, node := range nodeList {
+		if r.MatchString(node.Name()) {
+			mux.HandleFunc(fmt.Sprintf("/%s", strings.Split(node.Name(), ".")[0]), a.uiFiles)
+		}
+	}
```
shipperizer, line 29: just to keep the `RegisterEndpoints` lighter, split this in a separte private func

```
func (a *API) uiRoutes() []string {
    routes := make([]string, 0)
    
        nodeList, _ := fs.ReadDir(a.fileSystem, ".")
    	for _, node := range nodeList {
		if r.MatchString(node.Name()) {
			routes = append(routes, fmt.Sprintf("/%s", strings.Split(node.Name(), ".")[0]))
		}
	}
	return routes
}
```

or something similar

pkg/ui/handlers.go
bencekov: fixed

bencekov: I compiled the regex because we use it in a loop, so we match it with every file in ui/dist.

pkg/ui/handlers.go
```diff
@@ -1,23 +1,47 @@
...
+		mux.HandleFunc(route, a.uiFiles)
+	}
+}
+
+func (a *API) uiRoutes() []string {
+	routes := make([]string, 0)
+	nodeList, _ := fs.ReadDir(a.fileSystem, ".")
+
+	regex_html, _ := regexp.Compile(`html$`)
+	regex_misc, _ := regexp.Compile(`(svg|txt)$`)
```
shipperizer, line 33: I would put this in the constructor

at the same time u can bundle these into one, everything u put inside the round brackets can be fetched via a match result so you could use it in the if statement to decide how to strip the route path

pkg/ui/handlers.go
```diff
@@ -32,6 +56,10 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
 func NewAPI(fileSystem fs.FS, logger logging.LoggerInterface) *API {
 	a := new(API)
 
+	a.fileSystem = fileSystem
+
+	a.servePattern, _ = regexp.Compile(`(svg|txt|html)$`)
```
nsklikas, line 61: I am a little concerned that if someone wants to add a new file to the UI they will also have to change this.

nsklikas (commented):
I would like to point out that the UI and the backend will have different 404 pages/responses. This approach does not let the UI to do that.

Wouldn't it be better to move the UI under a subpatch (e.g. `/ui`) and use the file server as we did  before (but now the pattern would be `/ui/*`). I understand that it's not nice to see the in the URL the `ui` part, but we could handle that through the ingress (not sure though if the ingress relation allows for that).

Another approach would be to serve all files that have a file type using the file server (which means that you would add all the html files like you do now but also match this regex pattern: `*.*`). I think that this is a fair assumption to make as this is the file server and it makes sense for a file server to serve files, while an api server will never serve files which means that its paths will never end with a `.json` or `.something`.

bencekov:
> I would like to point out that the UI and the backend will have different 404 pages/responses. This approach does not let the UI to do that.
> 
> Wouldn't it be better to move the UI under a subpatch (e.g. `/ui`) and use the file server as we did before (but now the pattern would be `/ui/*`). I understand that it's not nice to see the in the URL the `ui` part, but we could handle that through the ingress (not sure though if the ingress relation allows for that).
> 
> Another approach would be to serve all files that have a file type using the file server (which means that you would add all the html files like you do now but also match this regex pattern: `*.*`). I think that this is a fair assumption to make as this is the file server and it makes sense for a file server to serve files, while an api server will never serve files which means that its paths will never end with a `.json` or `.something`.

shipperizer:
@nsklikas the `/ui/*` solution is the best one imho

pkg/ui/handlers.go
```diff
@@ -1,37 +1,80 @@
...
+	}
+
+	if a.isValidHtml(r.URL.Path) {
 		r.URL.Path += ".html"
 	}
 
+	url_parts := strings.Split(r.URL.Path, "/")
+	// strip /ui from url path
+	r.URL.Path = fmt.Sprintf("/%s", strings.Join(url_parts[2:], "/"))
+
```
shipperizer, line 65: i'm a bit lost in what happens here, as we keep changing the `r.URL.Path`

anyway after the consideration by @nsklikas about the `/ui` i guess we could carelessly serve all and just strip the `ui` prefix

not sure if we should add the `.html` suffix as well

pkg/ui/handlers.go
bencekov: ok

pkg/ui/handlers.go
bencekov: So what you see here, is that I don't want to serve routes such as /ui/login.html. I'd like to serve routes like /ui/login. At lines 62 to 64, we just take out the /ui part of the url path so that we can serve the requested page or file from the fileserver.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,80 @@
...
+	for _, node := range nodeList {
+		if match := a.servePattern.FindString(node.Name()); match == ".html" {
+			route := node.Name()[:len(node.Name())-len(match)]
+			routes = append(routes, fmt.Sprintf("/ui/%s", route))
+			continue
+		}
+	}
+	return routes
+}
```
nsklikas, line 40: IMO this is a little complicated and adds little value to the code base. 

If the only reason we do this is to remove the `.html` from the prometheus labels I think that it makes most sense to address it in the Prometheus middleware.

nsklikas (commented):
I have a couple of things that I don't like in this PR, but most of my comments are suggestions so feel free to argue or adopt them with changes.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,80 @@
...
+	return false
 }
 
 // TODO: Validate response when server error handling is implemented
 func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
-	if ext := path.Ext(r.URL.Path); ext == "" && r.URL.Path != "/" {
+
+	if match := a.servePattern.FindString(r.URL.Path); match == ".html" {
+		r.URL.Path = "/ui/404"
+	}
```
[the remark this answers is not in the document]
bencekov: I agree, I removed it.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,80 @@
...
+
+	if a.isValidHtml(r.URL.Path) {
 		r.URL.Path += ".html"
 	}
 
+	url_parts := strings.Split(r.URL.Path, "/")
+	// strip /ui from url path
+	r.URL.Path = fmt.Sprintf("/%s", strings.Join(url_parts[2:], "/"))
```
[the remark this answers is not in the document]
bencekov: done

pkg/ui/handlers.go
bencekov: ok, I removed it.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,55 @@
...
-		r.URL.Path += ".html"
+
+	if match := a.servePattern.FindString(r.URL.Path); match == ".html" {
+		w.WriteHeader(http.StatusNotFound)
+		// To make the output consistent with the fileserver
+		w.Write([]byte(fmt.Sprintf("%v page not found\n", http.StatusNotFound)))
+		return
+	}
+
+	if match := a.servePattern.FindString(r.URL.Path); match == "" {
```
nsklikas, line 37: Why use a regex and not `path.Ext` like we did before (it is designed for this, see https://pkg.go.dev/path#Ext)

pkg/ui/handlers.go
```diff
@@ -1,37 +1,55 @@
...
 func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
-	if ext := path.Ext(r.URL.Path); ext == "" && r.URL.Path != "/" {
-		r.URL.Path += ".html"
+
+	if match := a.servePattern.FindString(r.URL.Path); match == ".html" {
+		w.WriteHeader(http.StatusNotFound)
+		// To make the output consistent with the fileserver
+		w.Write([]byte(fmt.Sprintf("%v page not found\n", http.StatusNotFound)))
+		return
+	}
```
nsklikas, line 35: So we return 404 even though the file may exist on the server? That seems like weird behavior, I think we should remove this.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,55 @@
...
+
+	if match := a.servePattern.FindString(r.URL.Path); match == "" {
+		r.URL.Path = fmt.Sprintf("%s.html", r.URL.Path)
 	}
 
+	r.URL.Path = strings.TrimPrefix(r.URL.Path, "/ui")
```
nsklikas, line 41: `/ui` is used in 2 parts of the code, let's move it to a `const`.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,55 @@
...
+	fileServer   http.Handler
+	fileSystem   fs.FS
+	servePattern *regexp.Regexp
+	logger       logging.LoggerInterface
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	// TODO @shipperizer unsure if we deal with any POST/PUT/PATCH via js
-	mux.HandleFunc("/*", a.uiFiles)
+	mux.HandleFunc("/ui/*", a.uiFiles)
```
nsklikas, line 23: `/ui` is used in 2 parts of the code, let's move it to a `const`.

pkg/ui/handlers.go
```diff
@@ -1,37 +1,55 @@
...
 
+	r.URL.Path = strings.TrimPrefix(r.URL.Path, "/ui")
+
 	a.fileServer.ServeHTTP(w, r)
 }
 
 func NewAPI(fileSystem fs.FS, logger logging.LoggerInterface) *API {
 	a := new(API)
 
+	a.fileSystem = fileSystem
```
nsklikas, line 49: Is this used somewhere?

bencekov: taken out

pkg/ui/handlers.go
bencekov: done

pkg/ui/handlers.go
bencekov: done

pkg/ui/handlers.go
bencekov: sure

shipperizer (approved):

PR #113, opened 2023-08-15
IAM-387-Instrument Deep Health Check

bencekov:
IAM-387: instrumentation for components health check

This pr implements instrumentation for the deep health check.
After calling the /api/v0/deepcheck endpoint, you should see the results in the metrics. The metrics are available at /api/v0/metrics endpoint. The metric is called dependency_available.

### DIFF
--- a/internal/monitoring/interfaces.go
+++ b/internal/monitoring/interfaces.go
@@ -2,9 +2,6 @@ package monitoring
 
 type MonitorInterface interface {
 	GetService() string
-	GetResponseTimeMetric(map[string]string) (MetricInterface, error)
-}
-
-type MetricInterface interface {
-	Observe(float64)
+	SetResponseTimeMetric(map[string]string, float64) error
+	SetDependencyAvailability(map[string]string, float64) error
 }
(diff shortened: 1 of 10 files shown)


internal/monitoring/prometheus/prometheus.go
```diff
@@ -28,6 +30,19 @@ func (m *Monitor) GetResponseTimeMetric(tags map[string]string) (monitoring.Metr
 	return m.responseTime.With(tags), nil
 }
 
+func (m *Monitor) SetAvailability(component string, value float64) error {
```
shipperizer, line 33: I would stick with the pattern above used for the responseTime metric

```suggestion
func (m *Monitor) GetDependencyAvailability(tags map[string]string) (monitoring.MetricInterface, error)
```

internal/monitoring/interfaces.go
```diff
@@ -3,6 +3,7 @@ package monitoring
 type MonitorInterface interface {
 	GetService() string
 	GetResponseTimeMetric(map[string]string) (MetricInterface, error)
+	SetAvailability(string, float64) error
```
shipperizer, line 6: given gauges and histograms implement different methods best thing is to wrap each real metric into a struct, and if type is gauge observe method would panic or do nothing, if instead it s an histogram then the set method would behave like that

shipperizer (approved):

PR #117, opened 2023-08-18
Cardinality reduction for Prometheus metrics

bencekov:
In the current implementation, all requested route results in a separate http_response_time_seconds metric. This PR implements replacing the value of the route label of metrics belonging to requests with an invalid endpoint with unknown.

### DIFF
--- a/internal/monitoring/middlewares.go
+++ b/internal/monitoring/middlewares.go
@@ -3,7 +3,9 @@ package monitoring
 import (
 	"fmt"
 	"net/http"
+	"path"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/canonical/identity-platform-login-ui/internal/logging"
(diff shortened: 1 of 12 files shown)


shipperizer (commented):
there is a chi function named `Walk` (see example here https://github.com/go-chi/chi/issues/332) 

if the aim of this is to drop all the not found endpoints and replace them with "unknown" I would probably go for that, get a walk function to store all the routes and then simply check if current path is in the data structure

Only caveat is that middlewares are setup.before routes so you would have to call that after registering all the endpoints and update the middleware (and the respective data structure)

something like this https://github.com/canonical/identity-platform-login-ui/compare/main...v1/IAM-418

internal/monitoring/middlewares.go
```diff
@@ -14,6 +16,8 @@ const (
 	// IDPathRegex regexp used to swap the {id*} parameters in the path with simply id
 	// supports alphabetic characters and underscores, no dashes
 	IDPathRegex string = "{[a-zA-Z_]*}"
+	HTML        string = ".html"
+	UNKNOWN     string = "unknown"
```
nsklikas, line 20: I think that this should be something more descriptive like (is `view` a better word?)
```suggestion
	UNKNOWN     string = "<unknown endpoint>"
```

nsklikas:
My 2 cents:
- @shipperizer approach seems simpler.
- We should never hard code the UI pages in the backend code. If you wanted to do something like that I would expect that you would look into the fs.
- I don't really see the value around the `html` suffix changes. We are never going to have that many HTML pages that will cause cardinality problems.

PR #208, opened 2023-11-02
Openfga integration

nsklikas:
IAM-545

- Filter the providers based on the client name by calling openFGA
- Add docker-compose file for running all dependencies

Notes:
- If any of `apiScheme`, `apiHost`, `storeID`, `apiToken` are not defined, the noop authorization will be used (it will not filter the providers.
- If an app has no providers configured in openfga OR the providers in openfga are named differently from the ones on Kratos (e.g. typo), then a user will not be able to log in.
- We filter the provider list before we present it to the user
- We validate the provider the user chose
- We do not validate the provider after the user returns from kratos (successfully logs in)
- Both the http and grpc openfga APIs work.

To test this:
- You need a registered client with azure AD. You will need to create a `.env` file on the root of your repo, in there define the following vars:
  ```
  CLIENT_ID=<client-id>
  CLIENT_SECRET=<client-secret>
  MICROSOFT_TENANT=<tenant-id>
  ```
- Run `docker-compose -f docker-compose.dev.yml up --build --force-recreate` to launch all dependencies
- Start the login UI with the following env vars (on top of any other config you are already using):
  ```
  "OPENFGA_API_SCHEME": "http"
  "OPENFGA_API_HOST": "localhost:8081"
  "OPENFGA_STORE_ID": "01GP1254CHWJC1MNGVB0WDG1T0" (this is hardcoded in the docker-compose file)
  "OPENFGA_API_TOKEN": "42" (this is hardcoded in the docker-compose file)
  ```
- Use the following commands to create an entry for an app called `grafana` to access a provider named `microsoft`:
  ```
  export STORE_ID="01GP1254CHWJC1MNGVB0WDG1T0"
  fga model write --store-id $STORE_ID --file ./internal/authorization/schema.openfga
  export MODEL_ID=`fga model list --store-id $STORE_ID | yq .authorization_models.0.id`
  fga tuple write  --api-url http://localhost:8080 --store-id $STORE_ID --model-id $MODEL_ID app:grafana member app_group:cos
  fga tuple write  --api-url http://localhost:8080 --store-id $STORE_ID --model-id $MODEL_ID app_group:cos#member member app_group:everyone
  fga tuple write  --api-url http://localhost:8080 --store-id $STORE_ID --model-id $MODEL_ID app_group:everyone#member allowed_access provider:microsoft
  fga query list-objects  --api-url http://localhost:8080 --store-id $STORE_ID --model-id $MODEL_ID app:grafana allowed_access provider
  ```
- As an oidc client you can use the hydra CLI to test the flow. Run the following commands from the root of the repo:
  ```
  code_client=$(hydra create client \
    --endpoint http://localhost:4445 \
    --name grafana \
    --grant-type authorization_code,refresh_token \
    --response-type code \
    --format json \
    --scope openid,offline_access,email,profile \
    --redirect-uri http://127.0.0.1:4446/callback)
  hydra perform authorization-code   \
    -e http://localhost:4444 \
    --client-id `echo "$code_client" | yq .client_id` \
    --client-secret  `echo "$code_client" | yq .client_secret` \
    --scope openid,email,profile
  ```

TODO:
* [x] If openfga is not configured, disable authz
* [x] Add tests
* [x] Clean up logic
* [ ] UI should handle errors more gracefully


internal/authorization/authorization.go
```diff
@@ -0,0 +1,40 @@
...
+
+const (
+	DriverNoop    string = "noop"
+	DriverOpenFGA string = "openfga"
+)
+
+var authorizers = map[string]func() Authorizer{
+	DriverNoop:    func() Authorizer { return &NoopAuthorizer{} },
+	DriverOpenFGA: func() Authorizer { return &OpenFGAAuthorizer{} },
+}
```
shipperizer, line 18: have a look at https://go.dev/play/p/5V3KqjuVKTp

basically using a custom type definition and then define consts with well defined names

internal/authorization/noop_authorizer.go
```diff
@@ -0,0 +1,21 @@
+package authorization
+
+import (
+	"context"
+
+	"github.com/canonical/identity-platform-login-ui/internal/logging"
+)
+
+type NoopAuthorizer struct{}
```
shipperizer, line 9:
```suggestion
func NewNoopAuthorizer() *NoopAuthorizer {
  return new(NoopAuthorizer)
}

type NoopAuthorizer struct{}
```

internal/authorization/noop_authorizer.go
```diff
@@ -0,0 +1,21 @@
...
+	return objs, nil
+}
+
+func (a *NoopAuthorizer) Check(ctx context.Context, user string, relation string, object string) (bool, error) {
+	return true, nil
+}
+
+func (a *NoopAuthorizer) init(apiScheme, apiHost, storeID, apiToken, authModelID string, debug bool, logger logging.LoggerInterface) error {
+	return nil
+}
```
shipperizer, line 21:
```suggestion
```

internal/authorization/openfga_authorizer.go
```diff
@@ -0,0 +1,140 @@
...
+	return check.GetAllowed(), nil
+}
+
+func contains(s []string, e string) bool {
+	for _, a := range s {
+		if a == e {
+			return true
+		}
+	}
+	return false
```
shipperizer, line 139: lol, beauty of a DIY language

internal/authorization/authorization.go
```diff
@@ -0,0 +1,40 @@
...
+
+type Authorizer interface {
+	FilterObjects(context.Context, string, string, string, []string) ([]string, error)
+	Check(context.Context, string, string, string) (bool, error)
+	init(string, string, string, string, string, bool, logging.LoggerInterface) error
+}
+
+// Init the authorizer based on the config
+func NewAuthorizer(apiScheme, apiHost, storeID, apiToken, authModelID string, debug bool, logger logging.LoggerInterface) (*Authorizer, error) {
```
shipperizer, line 27: so, sticky point here, i'd separate the low level client with a NewClient like we do in the other internals, and then have a NewAuthorizer that accepts a client pointer, based on that it's a Noop or not, and there u ahve all the high level operations u need

internal/authorization/openfga_authorizer.go
```diff
@@ -0,0 +1,140 @@
...
+		return "", err
+	}
+
+	data, err := fgaClient.WriteAuthorizationModel(ctx).Body(builtinAuthorizationModel).Execute()
+	if err != nil {
+		return "", err
+	}
+
+	return data.GetAuthorizationModelId(), nil
+}
```
shipperizer, line 50: i'd chuck this on another object as it wont be really used by this codebase until we have a separate `app migrate` command but at that point we might want to separate the logic too

internal/authorization/openfga_authorizer.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (a *OpenFGAAuthorizer) init(apiScheme, apiHost, storeID, apiToken, authModelID string, debug bool, logger logging.LoggerInterface) error {
+	// TODO (nsklikas): Never create the model, we need to create a cli command for creating the model.
+	if authModelID == "" {
+		var err error
+		authModelID, err = WriteOpenFGAAuthorizationModel(context.Background(), apiScheme, apiHost, storeID)
+		if err != nil {
+			return err
+		}
+	}
```
shipperizer, line 60: ---> https://github.com/canonical/identity-platform-login-ui/pull/208/files#diff-a937587184869e26936fdd7775fbf8fdfa2d1f28c1eabfa3ad26b699ff28565aR53

pkg/kratos/handlers.go
```diff
@@ -30,14 +30,15 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
...
+	loginChallenge := q.Get("login_challenge")
 
 	// We try to see if the user is logged in, because if they are the CreateBrowserLoginFlow
 	// call will return an empty response
 	// TODO: We need to send a different content-type to CreateBrowserLoginFlow in order
 	// to avoid this bug.
 	session, _, _ := a.service.CheckSession(context.Background(), r.Cookies())
 	if session != nil {
-		redirectTo, cookies, err := a.service.AcceptLoginRequest(context.Background(), session.Identity.Id, q.Get("login_challenge"))
+		redirectTo, cookies, err := a.service.AcceptLoginRequest(context.Background(), session.Identity.Id, loginChallenge)
```
shipperizer, line 41: might wanna do the same change here as well 

https://github.com/canonical/identity-platform-login-ui/pull/208/files#diff-12151c8a9259cc64c5b82a9a51ed3eb6367a4850988ecd45cb323b4e086c708cL70
https://github.com/canonical/identity-platform-login-ui/pull/208/files#diff-12151c8a9259cc64c5b82a9a51ed3eb6367a4850988ecd45cb323b4e086c708cL75

pkg/kratos/handlers.go
```diff
@@ -79,6 +80,18 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
...
+		a.logger.Errorf("Error when filtering providers: %v\n", err)
+		switch err {
+		// If no providers are allowed, we simply log the error and return an empty flow
+		case ErrNoProvidersAllowed:
+		default:
+			http.Error(w, "Unexpected error", http.StatusInternalServerError)
+			return
+		}
+	}
+
```
shipperizer, line 94: this seems a bit odd, i know but cant make it any better as struggling to get my head around it but https://github.com/canonical/identity-platform-login-ui/pull/208/files#diff-5e4ba31969008b305d67765d5d4c9bd50d9bcc49571f1594e97d2b152ddf1378R27 doesn't look right and this code flow goes around that concept....not sure

pkg/kratos/service.go
```diff
@@ -163,6 +167,64 @@ func (s *Service) GetFlowError(ctx context.Context, id string) (*kClient.FlowErr
...
+	defer span.End()
+
+	provider := updateFlowBody.UpdateLoginFlowWithOidcMethod.Provider
+	clientName := loginFlow.Oauth2LoginRequest.Client.GetClientName()
+
+	allowed, err := s.authz.Check(ctx, fmt.Sprintf("app:%s", clientName), "allowed_access", fmt.Sprintf("provider:%s", provider))
+	if err != nil {
+		// We return false if an error occurs, this means that a fail on the authorization server will block our service
+		return false, err
+	}
```
shipperizer, line 181: i would expect that allowed is set to false if err is not nil and in that case a simple 

```
s.authz.Check(ctx, fmt.Sprintf("app:%s", clientName), "allowed_access", fmt.Sprintf("provider:%s", provider))
```

main reason is a bool defaults to false so it should be quite straightforward

pkg/kratos/service.go
```diff
@@ -163,6 +167,64 @@ func (s *Service) GetFlowError(ctx context.Context, id string) (*kClient.FlowErr
...
+
+	// This should happen only if the app has some allowed providers configured, but they don't match the ones on kratos
+	if len(allowedProviders) == 0 {
+		flow.Ui.Nodes = nil
+		return flow, ErrNoProvidersAllowed
+	}
+
+	// Filter UI nodes
+	var nodes []kClient.UiNode
+	for _, node := range flow.Ui.Nodes {
```
shipperizer, line 215: the double loop feels redundant, but to remove it we d need to get the FilterObjects to accept nodes and return nodes which right now seems overkill

still, sounds redundant

pkg/kratos/handlers.go
nsklikas: Since we don't treat it as an error, it probably should not be an error. It just feels weird from a UX perspective not returning any providers, so I added the error there in case we need to add some special logic in the future.

I will remove it

pkg/kratos/service.go
nsklikas: Yes I thought about that, but even in that case you would push the loops in the `FilterObjects`, I don't think we can avoid it.

internal/config/specs.go
```diff
@@ -17,6 +17,13 @@ type EnvSpec struct {
 
 	KratosPublicURL string `envconfig:"kratos_public_url"`
 	HydraAdminURL   string `envconfig:"hydra_admin_url"`
+
+	ApiScheme            string `envconfig:"openfga_api_scheme" default:""`
+	ApiHost              string `envconfig:"openfga_api_host"`
+	ApiToken             string `envconfig:"openfga_api_token"`
+	StoreId              string `envconfig:"openfga_store_id"`
+	AuthorizationModelId string `envconfig:"openfga_authorization_model_id" default:""`
+	AuthorizationEnabled bool   `envconfig:"authorization_enabled" default:"true"`
```
nsklikas, line 26: Perhaps `false` by default makes more sense?

cmd/main.go
```diff
@@ -65,7 +67,21 @@ func main() {
 	kClient := ik.NewClient(specs.KratosPublicURL, specs.Debug)
 	hClient := ih.NewClient(specs.HydraAdminURL, specs.Debug)
 
-	router := web.NewRouter(kClient, hClient, distFS, specs.BaseURL, tracer, monitor, logger)
+	var authzClient authz.AuthzClientInterface
+	if specs.AuthorizationEnabled {
+		logger.Info("Authorization is enabled")
+		cfg := fga.NewConfig(specs.ApiScheme, specs.ApiHost, specs.StoreId, specs.ApiToken, specs.AuthorizationModelId, specs.Debug)
+		authzClient = fga.NewClient(cfg)
+		if err != nil {
```
shipperizer, line 75: which err are we referring to?

internal/config/specs.go
shipperizer: not a biggie, whatever u deem more appropriate

internal/openfga/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+	check, _, err := a.APIClient().OpenFgaApi.CheckExecute(r)
+	if err != nil {
+		return false, err
+	}
+
+	return check.GetAllowed(), nil
+}
+
+func NewClient(cfg *Config) *Client {
```
shipperizer, line 108: given there is some substantial logic in here let's add logger, monitor and tracer

internal/openfga/noop.go
```diff
@@ -0,0 +1,15 @@
+package openfga
+
+import "context"
+
+type NoopClient struct{}
+
+func NewNoopClient() *NoopClient { return new(NoopClient) }
```
shipperizer, line 7: same as above for the magic 3, logger monitor and tracer

pkg/web/router.go
```diff
@@ -20,7 +21,7 @@ import (
 	"github.com/canonical/identity-platform-login-ui/pkg/ui"
 )
 
-func NewRouter(kratosClient *ik.Client, hydraClient *ih.Client, distFS fs.FS, baseURL string, tracer trace.Tracer, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) http.Handler {
+func NewRouter(kratosClient *ik.Client, hydraClient *ih.Client, authzClient *authz.Authorizer, distFS fs.FS, baseURL string, tracer trace.Tracer, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) http.Handler {
```
shipperizer, line 24: this needs to be the interface of the authorizer

pkg/kratos/service.go
```diff
@@ -207,3 +260,12 @@ func unmarshalByteJson(data io.Reader, v any) error {
 	}
 	return nil
 }
+
+func contains(s []string, e string) bool {
```
shipperizer, line 264: as wrong and bad as it sounds, make it a method

Makefile
```diff
@@ -43,4 +43,8 @@ cmd/ui/dist:
 
 npm-build:
 	$(MAKE) -C ui/ build
-.PHONY: npm-build
\ No newline at end of file
+.PHONY: npm-build
+
+update-openfga-model:
+	@printf 'package openfga\n\n// Code generated by Makefile; DO NOT EDIT.\n\nvar authModel = `%s`\n' '$(shell npx @openfga/syntax-transformer transform --from=dsl --inputFile=./internal/openfga/schema.openfga | jq -c)' > ./internal/openfga/auth_model.go
```
aaguiarz, line 49: Hey! I'm the product manager for OpenFGA and found this PR :)

You can do this using the [OpenFGA CLI](https://github.com/openfga/cli) using `fga model transform ./internal/openfga/schema.openfga`, and actually you don't need the JSON representation anymore, you can just write the `.openfga` model directly to a store using `fga model write --store-id=.. --file schema.openfga` without having code generated by the makefile.

internal/authorization/schema.openfga
```diff
@@ -0,0 +1,16 @@
+model
+  schema 1.1
+type user
+type app
+type group
```
aaguiarz, line 5: It seems you are trying to model nested groups here.

A simpler/more performant way to do it would be 

```
type group
   relations
       define member : [user, group#member]
type app_group
    relations
       define member : [app, app_group#member]

If you want to define that the group:engineering is a child of group:everyone, you'd write a tuple

USER: group:engineering#member
RELATION: member
OBJECT: group:everyone

aaguiarz (commented):
Hi! I'm the product manager for OpenFGA, found this PR and I'm suggesting a few improvements :)

Let me know if we can help in any way.

Makefile
nsklikas: Hello! Thanks a lot for the suggestions.

This is used for transforming the `.openfga` model to a go object. It is a hacky approach and we will probably have to revisit it.

internal/authorization/schema.openfga
nsklikas: Point taken, thanks for the suggestion

shipperizer (approved):

PR #211, opened 2023-12-05
Device flow POC

nsklikas:
POC for Hydra device flow.

To test:
- Bring up the rest of the components with: 
```console
docker-compose -f docker-compose.dev.yml up --build --force-recreate
```
- Register a client in Hydra that can:
```console
hydra create client \
  --endpoint http://127.0.0.1:4445 \
  --name noauth \
  --grant-type urn:ietf:params:oauth:grant-type:device_code \
  --scope openid,offline \
  --token-endpoint-auth-method none
```
- You can use the client found in https://github.com/canonical/hydra-rock/tree/IAM-597 (only config needed is `OAUTH_CLIENT_ID="<client_id>"`, where <client_id> is taken from the previous step)
- Run the login UI with the usual config

Note
- [ ]  I couldn't find a way to fetch info about the client so that we can ask for the user to authorize that specific device.

### DIFF
--- a/ui/pages/device.tsx
+++ b/ui/pages/device.tsx
@@ -1,0 +1,73 @@
+import { Card, Row } from "@canonical/react-components";
+import axios from "axios";
+import Head from "next/head";
+import { useRouter } from "next/router";
+import React, { FormEvent, useCallback, useEffect } from "react";
+
+export interface Response {
+  data: {
+    redirect_to?: string;
+  };
+}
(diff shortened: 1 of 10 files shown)


ui/pages/device.tsx
edlerd, line 1: Can be built on top of `/ui/pages/device_code.tsx` which comes with new layout.

PR #215, opened 2024-01-05
Add create-fga-model command

nsklikas:
IAM-631

This PR:
- Uses cobra to create the login UI CLI
- Introduces 3 commands: `serve`, `version`, `create-fga-model`

This is a breaking change:
- The command that runs the server is now `binary serve` instead of `binary`
- The command that returns the version is now `binary version` instead of `binary -v`

image


Makefile
```diff
@@ -31,7 +31,7 @@ vendor:
 .PHONY: vendor
 
 build: cmd/ui/dist
-	$(MAKE) -C cmd build
+	$(GO) build -o $(GO_BIN) ./
```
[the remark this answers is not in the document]
shipperizer: ah https://github.com/canonical/identity-platform-login-ui/pull/215/files#diff-67c464bbed217d3d422ddd2db32d89360c18ce8584542858e61f3f842c6fb432R39
either u beat me to it, or the error is somewhere else

internal/tracing/interfaces.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"context"
+
+	"go.opentelemetry.io/otel/trace"
+)
+
+type TracingInterface interface {
+	Start(context.Context, string, ...trace.SpanStartOption) (context.Context, trace.Span)
+}
```
shipperizer, line 11: if this is done becuase of the update of otel lib to `v1.21.0` then worth looking also at this implementation https://github.com/glauth/glauth/commit/1a3739610997b58100040d0c8a405596fccc8e23

now if this works i m happy with it, we can always review it down the line as i dunno the implications of not using the standard `Tracer`, at the same time if we only need `Start`, as we seem to be, all is good

cmd/createFgaModel.go
```diff
@@ -0,0 +1,67 @@
...
+	tracer := tracing.NewNoopTracer(nil)
+	monitor := monitoring.NewNoopMonitor("", logger)
+	scheme, host, err := parseURL(apiUrl)
+	if err != nil {
+		panic(err)
+	}
+	cfg := fga.NewConfig(scheme, host, storeId, apiToken, "", false, tracer, monitor, logger)
+	authzClient := fga.NewClient(cfg)
+	authorizer := authz.NewAuthorizer(authzClient, tracer, monitor, logger)
+	modelId, err := authorizer.CreateModel(context.Background())
```
shipperizer, line 54: instead of enhancing the Authorizer, i would probably put it all inline here

https://github.com/canonical/identity-platform-login-ui/pull/215/files#diff-37ac464c45e06bed5132281126e435b869de3ca34bbdfb008db79edad30c8edcR60-R61

although if reasoning is to keep the logic and model insync happy to be proven wrong and keep it as is

internal/logging/noop.go
```diff
@@ -0,0 +1,19 @@
+package logging
+
+type NoopLogger struct {
+}
+
+func NewNoopLogger() *NoopLogger {
```
shipperizer, line 6: why not using https://pkg.go.dev/go.uber.org/zap@v1.26.0#NewNop?

cmd/createFgaModel.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/spf13/cobra"
+)
+
+// createFgaModelCmd represents the createFgaModel command
+var createFgaModelCmd = &cobra.Command{
+	Use:   "create-fga-model",
+	Short: "Creates an openfga model",
+	Long:  `Creates an openfga model`,
+	Run: func(cmd *cobra.Command, args []string) {
+		apiUrl, _ := cmd.Flags().GetString("api-url")
```
shipperizer, line 25: `app create-fga-model --api-url ...` 
even though there is the `fga` bit, not sure `api-url` is enough to explain what url is needed
we might need to either change the name of the command and simplify the args or the other way round

anyway not a biggie

cmd/createFgaModel.go
nsklikas: The reason is to use the model, which is at https://github.com/canonical/identity-platform-login-ui/blob/main/internal/authorization/auth_model.go. I could make the variable global and use that.

internal/logging/noop.go
nsklikas: Good catch, will change it

internal/tracing/interfaces.go
nsklikas: I added it to keep it consistent with the [monitoring](https://github.com/canonical/identity-platform-login-ui/blob/main/internal/monitoring/interfaces.go#L3) and [logging](https://github.com/canonical/identity-platform-login-ui/blob/main/internal/logging/interfaces.go#L3). No problem with removing it if you think it is out of place

cmd/createFgaModel.go
```diff
@@ -0,0 +1,67 @@
...
+	createFgaModelCmd.PersistentFlags().String("api-url", "", "The openfga API URL")
+	createFgaModelCmd.PersistentFlags().String("api-token", "", "The openfga API token")
+	createFgaModelCmd.PersistentFlags().String("store-id", "", "The openfga store to create the model in")
+	createFgaModelCmd.MarkPersistentFlagRequired("api-url")
+	createFgaModelCmd.MarkPersistentFlagRequired("api-token")
+	createFgaModelCmd.MarkPersistentFlagRequired("store-id")
+}
+
+func createModel(apiUrl, apiToken, storeId string) {
+	logger := logging.NewNoopLogger()
```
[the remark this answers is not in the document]
nsklikas: The reason that i didn't use the logger is that we don't want any of the features the logger provides here. The logger is configured to write to a file and add a timestamp along with other sugar-coating. IMO for a CLI command all we want is to print to the CLI, so even if we didn't forward the logs to a file it would not be needed. 

I thought about adding a `plain` logger struct, but I decided against it because I believe that these messages aren't exactly "logs". They are a command's output. Not sure if makes sense

shipperizer (approved):

PR #216, opened 2024-01-26
Use new design and add user flows as dummy pages WD-8469

edlerd:
* adjusted to new design
* added dummy pages for register, reset, device and setup


rockcraft.yaml
```diff
@@ -1,7 +1,7 @@
 name: identity-platform-login-ui
 
 base: bare
-build-base: ubuntu:22.04
+build-base: ubuntu@22.04
```
shipperizer, line 4: can we separate this into another commit?
or simply drop it and we ll chuck it in the main branch and u can rebase

shipperizer (commented):
it is quite a lot of code in just one commit, any chance to split it into smaller ones?

edlerd:
> it is quite a lot of code in just one commit, any chance to split it into smaller ones?

It's not that many code changes, most of the lines are due to dependency updates, which are a prerequisite for other changes. So factoring them out into individual PRs is painful, because they would rely on each other.

ui/pages/login.tsx
```diff
@@ -96,27 +96,14 @@ const Login: NextPage = () => {
 
           return Promise.reject(err);
         }),
-    [flow, router]
+    [flow, router],
   );
+  const reqDomain = flow?.return_to ? new URL(flow.return_to).hostname : "";
```
nsklikas, line 101: This is not correct. This `return_to` always points to Hydra, to get info about the client you would have to look into `flow.oauth2_login_request.client`. There are 2 fields there that we can use `client_name` which is supposed to be the human readable name of the client and `client_uri` which is the client's homepage (see https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata for more info)

edlerd: Good point, thanks for the clarification. I changed this to construct a string with client name and client url if present.

rockcraft.yaml
edlerd: I factored this into a dedicated commit.

nsklikas (commented):
LGTM other than the comment I made on the login page. One thing I don't understand is what the plan is for these registration pages, the fields are going to come from Kratos and I would expect that we are going to ask for more attributes rather than just username and password. We can iterate on this further when we do the work on the backend so I have no problem with merging this once the issue on the login page is fixed.

Here are some screen shots on what the new pages look like:
Login Page:
image

OIDC Error:
image

Register Secure:
image

Register Complete:
image

Register Email:
image

Register Password:
image

Register Secure:
image

Reset Check:
image

Reset Complete:
image

Reset Email:
image

Reset Password:
image

Setup Complete:
image

Setup Password:
image

Setup Secure:
image

shipperizer:
@edlerd looks good to me afaict, what is the policy for frontend tests?how do we catch broken UI?

ui/components/NodeInputSubmit.tsx
```diff
@@ -9,15 +9,59 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
+    }
+    if (value.toLowerCase().startsWith("google")) {
+      return "logos/Google.svg";
+    }
+    if (value.toLowerCase().startsWith("microsoft")) {
+      return "logos/Microsoft.svg";
+    }
+    if (value.toLowerCase().startsWith("okta")) {
+      return "logos/Okta.svg";
+    }
```
nsklikas, line 28: There is no Okta provider type in Kratos, let's remove this and the Okta icon logo  (see https://www.ory.sh/docs/kratos/social-signin/overview)

ui/public/logos/Okta.svg
nsklikas, line 1: There is not Okta provider type in Kratos, let's remove this (see https://www.ory.sh/docs/kratos/social-signin/overview)

nsklikas (commented):
One more minor thing and I think we can merge it

ui/components/NodeInputSubmit.tsx
edlerd: removed the entry for okta.

ui/public/logos/Okta.svg
edlerd: removed this asset

nsklikas (approved):
LGTM

shipperizer (approved):

lukasSerelis:
Looks good. MFA options obviously missing but I assume this is out of scope at this stage

PR #222, opened 2024-02-05
Iam 671 fix `CheckAllowedProvider` function

### LINKED ISSUE #221: Login UI requires openfga deployment
Using the latest image, login ui requires openfga deployment in order to work, otherwise it fails to return the allowed identity providers even if `AUTHORIZATION_ENABLED=false`.
```
poc-kratos-selfservice-ui-node-1  | 2024-02-01T14:27:46.199Z [login-ui] github.com/canonical/identity-platform-login-ui/pkg/kratos.(*Service).CheckAllowedProvider(0xc000280720, {0xf98500?, 0x1599f00?}, 0xc000150000, 0xc00021aa80)
poc-kratos-selfservice-ui-node-1  | 2024-02-01T14:27:46.199Z [login-ui] 	/root/parts/go-build/build/pkg/kratos/service.go:173
```
To reproduce:
Set up a docker environment with kratos, hydra and login ui (for example https://github.com/natalian98/oathkeeper-poc/tree/test-with-httpbin) pinning the login ui version to latest

BarcoMasile:
# Description
This PR fixes the `CheckAllowedProvider` for the OAuthkeeper case in which the `Oauth2LoginRequest` from the `loginFlow` object would be missing since OAuthkeeper doesn't perform OAuth2 requests.

A new simple function has been introduced with tests.
This PR closes #221.

### Side notes
- `log.txt` added to the `.gitignore` file
- fixed example value from the doc


pkg/kratos/service.go
```diff
@@ -183,12 +183,20 @@ func (s *Service) CheckAllowedProvider(ctx context.Context, loginFlow *kClient.L
 	return s.contains(allowedProviders, fmt.Sprintf("%v", provider)), nil
 }
 
+func getClientName(loginFlow *kClient.LoginFlow) string {
```
shipperizer, line 186: I would make it a method even if Service attributes are not used

it s simply a helper function so if it has to be separate then better to be part of the struct
other option is to keep it inline which leads to repetition

ui/package-lock.json
```diff
@@ -33,7 +33,7 @@
         "typescript": "5.3.3"
       },
       "engines": {
-        "node": "18"
+        "node": "20"
```
shipperizer, line 36: Might wanna drop this change

pkg/kratos/service.go
BarcoMasile: Thanks, I'll go the method route!

ui/package-lock.json
BarcoMasile: I'll drop it then, this was just to align with package.json

pkg/kratos/service_test.go
```diff
@@ -731,6 +732,26 @@ func TestCheckAllowedProviderFail(t *testing.T) {
...
 }
 
+func TestGetClientNameOAuthKeeper(t *testing.T) {
+	loginFlow := &kClient.LoginFlow{}
+	service := NewService(nil, nil, nil, nil, nil, nil)
+
+	actualClientName := service.GetClientName(loginFlow)
+
+	const expectedClientName = ""
+	assert.Equal(t, expectedClientName, actualClientName)
```
nsklikas, line 742: In our other tests we don't use this library, let's keep it consistent

pkg/kratos/service.go
```diff
@@ -183,12 +183,20 @@ func (s *Service) CheckAllowedProvider(ctx context.Context, loginFlow *kClient.L
 	return s.contains(allowedProviders, fmt.Sprintf("%v", provider)), nil
 }
 
+func (s *Service) GetClientName(loginFlow *kClient.LoginFlow) string {
```
shipperizer, line 186: It's only used internally so u can keep it private and lowercase

not a blocker anyway

shipperizer (approved):

pkg/kratos/service.go
BarcoMasile: I didn't realize you could still use unexported field in the same package. Thanks!
Done in new revision!

pkg/kratos/service_test.go
```diff
@@ -731,6 +731,30 @@ func TestCheckAllowedProviderFail(t *testing.T) {
 	}
 }
 
+func TestGetClientNameOAuthKeeper(t *testing.T) {
```
natalian98, line 734: small nitpick: it's `oathkeeper` :)

nsklikas (approved):

PR #226, opened 2024-03-29
Update device code page

nsklikas:
IAM-736:
- Update the device code page.
image

There are a couple problems with this PR:
- We do not use the Hydra sdk to call the Hydra API. If we wish to use the client sdk do we have to fork https://github.com/ory/hydra-client-go?
- It does not handle errors from the backend. If the user inputs a wrong code, then nothing happens
- From a UX perspective it would be better if we had one box per letter rather than a single box for the whole code, e.g. see netflix:
image

### DIFF
--- a/internal/hydra/client.go
+++ b/internal/hydra/client.go
@@ -3,28 +3,29 @@ package hydra
 import (
 	"net/http"
 
-	client "github.com/ory/hydra-client-go/v2"
+	hClient "github.com/ory/hydra-client-go/v2"
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
 type Client struct {
-	c *client.APIClient
+	c         *hClient.APIClient
+	deviceApi *DeviceApiService
 }
 
-func (c *Client) OAuth2Api() client.OAuth2Api {
-	return c.c.OAuth2Api
+func (c *Client) OAuth2Api() OAuth2Api {
+	return c.deviceApi
 }
 
-func (c *Client) MetadataApi() client.MetadataApi {
+func (c *Client) MetadataApi() hClient.MetadataApi {
 	return c.c.MetadataApi
 }
 
 func NewClient(url string, debug bool) *Client {
 	c := new(Client)
 
-	configuration := client.NewConfiguration()
+	configuration := hClient.NewConfiguration()
 	configuration.Debug = debug
-	configuration.Servers = []client.ServerConfiguration{
+	configuration.Servers = []hClient.ServerConfiguration{
 		{
 			URL: url,
 		},
@@ -32,7 +33,8 @@ func NewClient(url string, debug bool) *Client {
 
 	configuration.HTTPClient = &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
 
-	c.c = client.NewAPIClient(configuration)
+	c.c = hClient.NewAPIClient(configuration)
+	c.deviceApi = newDeviceApiService(c.c)
 
 	return c
 }
--- a/internal/hydra/device.go
+++ b/internal/hydra/device.go
@@ -1,0 +1,237 @@
+package hydra
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	hClient "github.com/ory/hydra-client-go/v2"
+)
+
+// We implement the device API logic, because the upstream sdk does not support it.
+// Otherwise we would have to fork the sdk
+// TODO(nsklikas): Remove this once upstream hydra supports the device flow
+
+type DeviceApiService struct {
+	client *hClient.APIClient
+	hClient.OAuth2Api
+}
+
+type APIError struct {
+	body  []byte
+	error string
+}
+
+// Error returns non-empty string if there was an error.
+func (e APIError) Error() string {
+	return e.error
+}
+
+// Body returns the raw bytes of the response
+func (e APIError) Body() []byte {
+	return e.body
+}
+
+// Prevent trying to import "fmt"
+func reportError(format string, a ...interface{}) error {
+	return fmt.Errorf(format, a...)
+}
+
+// AcceptDeviceUserCodeRequest Contains information on an device verification
+type AcceptDeviceUserCodeRequest struct {
+	UserCode *string `json:"user_code,omitempty"`
+}
+
+// NewAcceptDeviceUserCodeRequest instantiates a new AcceptDeviceUserCodeRequest object
+// This constructor will assign default values to properties that have it defined,
+// and makes sure properties required by API are set, but the set of arguments
+// will change when the set of required properties is changed
+func NewAcceptDeviceUserCodeRequest() *AcceptDeviceUserCodeRequest {
+	this := AcceptDeviceUserCodeRequest{}
+	return &this
+}
+
+// NewAcceptDeviceUserCodeRequestWithDefaults instantiates a new AcceptDeviceUserCodeRequest object
+// This constructor will only assign default values to properties that have it defined,
+// but it doesn't guarantee that properties required by API are set
+func NewAcceptDeviceUserCodeRequestWithDefaults() *AcceptDeviceUserCodeRequest {
+	this := AcceptDeviceUserCodeRequest{}
+	return &this
+}
+
+// GetUserCode returns the UserCode field value if set, zero value otherwise.
+func (o *AcceptDeviceUserCodeRequest) GetUserCode() string {
+	if o == nil || o.UserCode == nil {
+		var ret string
+		return ret
+	}
+	return *o.UserCode
+}
+
+// GetUserCodeOk returns a tuple with the UserCode field value if set, nil otherwise
+// and a boolean to check if the value has been set.
+func (o *AcceptDeviceUserCodeRequest) GetUserCodeOk() (*string, bool) {
+	if o == nil || o.UserCode == nil {
+		return nil, false
+	}
+	return o.UserCode, true
+}
+
+// HasUserCode returns a boolean if a field has been set.
+func (o *AcceptDeviceUserCodeRequest) HasUserCode() bool {
+	if o != nil && o.UserCode != nil {
+		return true
+	}
+
+	return false
+}
+
+// SetUserCode gets a reference to the given string and assigns it to the UserCode field.
+func (o *AcceptDeviceUserCodeRequest) SetUserCode(v string) {
+	o.UserCode = &v
+}
+
+func (o AcceptDeviceUserCodeRequest) MarshalJSON() ([]byte, error) {
+	toSerialize := map[string]interface{}{}
+	if o.UserCode != nil {
+		toSerialize["user_code"] = o.UserCode
+	}
+	return json.Marshal(toSerialize)
+}
+
+type NullableAcceptDeviceUserCodeRequest struct {
+	value *AcceptDeviceUserCodeRequest
+	isSet bool
+}
+
+func (v NullableAcceptDeviceUserCodeRequest) Get() *AcceptDeviceUserCodeRequest {
+	return v.value
+}
+
+func (v *NullableAcceptDeviceUserCodeRequest) Set(val *AcceptDeviceUserCodeRequest) {
+	v.value = val
+	v.isSet = true
+}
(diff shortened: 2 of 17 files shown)


shipperizer (dismissed):

internal/hydra/client.go
```diff
@@ -19,6 +20,13 @@ func (c *Client) MetadataApi() client.MetadataApi {
 	return c.c.MetadataApi
 }
 
+// We implement the device API logic, as the upstream sdk does not support it.
+// Otherwise we would have to fork the sdk
+// TODO(nsklikas): Remove this once upstream hydra supports the device flow
+func (c *Client) DeviceApi() DeviceApi {
+	return c.deviceApi
+}
```
shipperizer, line 28: had another thought, would we be able to embed these into the oauth2Api instead?

```
 func (c *Client) OAuth2Api() OAuth2Api {
.
.
.

type OAuth2Api interface {
   client.OAuth2Api
   AcceptUserCodeRequest(context.Context) ApiAcceptUserCodeRequestRequest
   AcceptUserCodeRequestExecute(ApiAcceptUserCodeRequestRequest) (*hClient.OAuth2RedirectTo, *http.Response, error)
}
``` 

we would wrap only the oauth2 client basically

internal/hydra/device.go
```diff
@@ -0,0 +1,210 @@
...
+// Otherwise we would have to fork the sdk
+// TODO(nsklikas): Remove this once upstream hydra supports the device flow
+
+type DeviceApi interface {
+	AcceptUserCodeRequest(context.Context) ApiAcceptUserCodeRequestRequest
+	AcceptUserCodeRequestExecute(ApiAcceptUserCodeRequestRequest) (*hClient.OAuth2RedirectTo, *http.Response, error)
+}
+
+type DeviceApiService struct {
+	client *hClient.APIClient
```
shipperizer, line 24: based on the change this would become https://github.com/canonical/identity-platform-login-ui/pull/226#discussion_r1548375587

shipperizer (commented):
see extra comments

lukasSerelis:
Addressed the no error state for incorrect device code [here](https://www.figma.com/file/eFR98XGAliMnPUWeBtWzKI/Login-UI?type=design&node-id=661-8632&mode=design&t=ZVuO8d3WP0gdQZi8-4)

Regarding the split of text field per character is low priority, also we don't have front-end components to support this atm. The placeholder text sort-of helps this, the goal is just to match the placeholder formatting to the console formatting

nsklikas:
Added error handling in the UI

the code is invalid
image

hydra is down, so we get an unexpected error
image

ui/pages/device_code.tsx
```diff
@@ -1,25 +1,83 @@
...
+  const router = useRouter();
+  const { device_challenge: challenge, user_code: code } = router.query;
+  const [errorMessage, setErrorMessage] = useState("");
+
+  useEffect(() => {
+    // If the router is not ready yet, or we already have a flow, do nothing.
+    if (!router.isReady) {
+      return;
+    }
+  }, [code, errorMessage]);
```
edlerd, line 36: This is a no-op, can it be removed? Or should the content maybe not be wrapped in a use effect?

ui/pages/device_code.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+export interface Response {
+  data: {
+    redirect_to?: string;
+  };
+}
+
+async function acceptUserCode(userCode: string, challenge: string) {
+  return axios
+    .put("/api/device?device_challenge=" + String(challenge), {
```
edlerd, line 16:
```suggestion
    .put(`/api/device?device_challenge=${challenge}`, {
```
We can use the inline style here and don't need to cast to string again, I guess?

ui/pages/device_code.tsx
```diff
@@ -1,25 +1,83 @@
...
+
+      const formData = new FormData(event.currentTarget);
+      const userCode = formData.get("code");
+      acceptUserCode(String(userCode), String(challenge)).catch(
+        (error: AxiosError) => {
+          // TODO(nsklikas): Refactor when proper error handling is
+          // implemented in the backend
+          if (error.response?.status == 400) {
+            setErrorMessage(
+              "The code is either invalid, expired or already used",
```
edlerd, line 50: Isn't the `error.response.data.error_description` available to be used here?

nsklikas: The backend will return an error saying `The user_code provided is either invalid, expired or already used` (it's not a json after all, to be consistent with the rest of the endpoints).

We do a very bad job at how we expose errors in the backed, we will need to refactor it in the future and we can revisit this then

ui/pages/device_code.tsx
nsklikas: I removed the `useEffect`, should be fine now

nsklikas:
@lukasSerelis @edlerd

Not sure if this makes sense, but there are 2 case when you end up in the `device_code` page, you either get there :
- with a URL without a `user_code` param, in which case the box will simply show a placeholder 
image
- with a URL with a `user_code` param, in which case the input box will be pre-populated with the value of the user_code.

Currently in the second case you will be unable to edit the user code (you can't change type a different code), not sure if this makes sense and I am not sure why this happens (probably because of https://github.com/canonical/identity-platform-login-ui/pull/226/files#diff-90614cb17aaa983f34c429b732b82f94716e1f4fa61c25c752f7c43d0d702537R63 @edlerd ?)

My question is: is this wrong? do we want to allow the users to edit the code? (probably yes I would say)

edlerd:
> @lukasSerelis @edlerd
> 
> Not sure if this makes sense, but there are 2 case when you end up in the `device_code` page, you either get there :
> 
> * with a URL without a `user_code` param, in which case the box will simply show a placeholder
>   image
> * with a URL with a `user_code` param, in which case the input box will be pre-populated with the value of the user_code.
> 
> Currently in the second case you will be unable to edit the user code (you can't change type a different code), not sure if this makes sense and I am not sure why this happens (probably because of https://github.com/canonical/identity-platform-login-ui/pull/226/files#diff-90614cb17aaa983f34c429b732b82f94716e1f4fa61c25c752f7c43d0d702537R63 @edlerd ?)
> 
> My question is: is this wrong? do we want to allow the users to edit the code? (probably yes I would say)

Shouldn't we auto submit the form if the user comes with a code already in the URL?

If we want the user to be able to change the code, better use `defaultValue={code}` instead of `value={code}`.

internal/hydra/device.go
```diff
@@ -0,0 +1,231 @@
...
+		return nil, nil, err
+	}
+	url := localBasePath + "/admin/oauth2/auth/requests/device/accept?challenge=" + *r.deviceChallenge
+	req, err := http.NewRequest(http.MethodPut, url, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, nil, reportError("error when constructing request: %s", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{}
```
shipperizer, line 197: is it worth using a shared client? ideally we could even wrap it with the `otel` http client to spread traces (which can help with troubleshooting)

internal/hydra/interfaces.go
```diff
@@ -0,0 +1,22 @@
...
+	hClient "github.com/ory/hydra-client-go/v2"
+)
+
+// We implement the device API logic, as the upstream sdk does not support it.
+// Otherwise we would have to fork the sdk.
+// TODO(nsklikas): Remove this once upstream hydra supports the device flow
+type OAuth2Api interface {
+	hClient.OAuth2Api
+	AcceptUserCodeRequest(context.Context) ApiAcceptUserCodeRequestRequest
+	AcceptUserCodeRequestExecute(ApiAcceptUserCodeRequestRequest) (*hClient.OAuth2RedirectTo, *http.Response, error)
```
shipperizer, line 16: you can reuse the interface below

internal/hydra/device.go
```diff
@@ -0,0 +1,231 @@
...
+	body, err := json.Marshal(r.acceptDeviceUserCodeRequest)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	localBasePath, err := a.client.GetConfig().ServerURLWithContext(r.ctx, "OAuth2ApiService.AcceptUserCodeRequest")
+	if err != nil {
+		return nil, nil, err
+	}
+	url := localBasePath + "/admin/oauth2/auth/requests/device/accept?challenge=" + *r.deviceChallenge
```
shipperizer, line 190: I guess.tjis follows ory sdk styling, right?

internal/hydra/device.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Prevent trying to import "fmt"
+func reportError(format string, a ...interface{}) error {
+	return fmt.Errorf(format, a...)
+}
+
+// AcceptDeviceUserCodeRequest Contains information on an device verification
+type AcceptDeviceUserCodeRequest struct {
+	UserCode *string `json:"user_code,omitempty"`
```
shipperizer, line 45: do we need validation here?

nsklikas:
> Shouldn't we auto submit the form if the user comes with a code already in the URL?

That was my initial thought as well, but the [RFC](https://datatracker.ietf.org/doc/html/rfc8628#section-3.3.1) says:

> If the user starts the user interaction by navigating to
   "verification_uri_complete", then the user interaction described in
   [Section 3.3](https://datatracker.ietf.org/doc/html/rfc8628#section-3.3) is still followed, with the optimization that the user
   does not need to type in the "user_code".  The server SHOULD display
   the "user_code" to the user and ask them to verify that it matches
   the "user_code" being displayed on the device to confirm they are
   authorizing the correct device.  As before, in addition to taking
   steps to confirm the identity of the device, the user should also be
   afforded the choice to approve or deny the authorization request.

And it makes sense from a security perspective, so that you are sure that you are validating your device and that you are not being scammed

internal/hydra/device.go
nsklikas: This struct eventually will be provided by the ory sdk, which does not provide validation. So I think we should leave it as it is and if we want for further validation we should implement it at another layer.

internal/hydra/device.go
nsklikas: This whole function is written by me, trying to mimic the ory sdk.

I will refactor this part a little as it is too ugly.

internal/hydra/device.go
nsklikas: Good point, I changed it

lukasSerelis:
> @lukasSerelis @edlerd
> 
> Not sure if this makes sense, but there are 2 case when you end up in the `device_code` page, you either get there :
> 
>     * with a URL without a `user_code` param, in which case the box will simply show a placeholder
>       image
> 
>     * with a URL with a `user_code` param, in which case the input box will be pre-populated with the value of the user_code.
> 
> 
> Currently in the second case you will be unable to edit the user code (you can't change type a different code), not sure if this makes sense and I am not sure why this happens (probably because of https://github.com/canonical/identity-platform-login-ui/pull/226/files#diff-90614cb17aaa983f34c429b732b82f94716e1f4fa61c25c752f7c43d0d702537R63 @edlerd ?)
> 
> My question is: is this wrong? do we want to allow the users to edit the code? (probably yes I would say)

I think having the code non editable is okay, but some surrounding copy would need to change. Instead of "Enter the code to access ..." it should say "Verify the code to access ..." or something between those lines. Ideally text field should signify that it's not editable as well, to not cause any confusion, and to avoid any edge edge edge cases we should add a way to go back to the empty state where they could enter a code manually. 
image

nsklikas:
> Instead of "Enter the code to access ..." it should say "Verify the code to access ..." or something between those lines

@lukasSerelis unfortunately we don't know what application the user is trying to access at this stage. The user_code is used to connect the login session with the device. This is why the rfc states that the user code verification should be followed by a consent for the device (which I think we are planning not to implement, at least for now)

> Ideally text field should signify that it's not editable as well, to not cause any confusion, and to avoid any edge edge edge cases we should add a way to go back to the empty state where they could enter a code manually.

I would like to merge this PR with out changing the UI design too much and to polish the UI on a later stage, preferably someone else with more UI knowledge should do it. Currently I made it editable and we can iterate on it.

edlerd (approved):
Typescript part LGTM

PR #227, opened 2024-04-11
Add success screen for device flow WD-10251

edlerd:
# Done

Added success screen for device flow under the url http://localhost:4455/ui/device_complete

Fixes: WD-10251

# Screenshot

image

### DIFF
--- a/ui/pages/device_complete.tsx
+++ b/ui/pages/device_complete.tsx
@@ -1,0 +1,15 @@
+import type { NextPage } from "next";
+import React from "react";
+import PageLayout from "../components/PageLayout";
+
+const DeviceComplete: NextPage = () => {
+  return (
+    <PageLayout title="Sign in successful">
+      <p>
+        Your device has been successfully connected. You can close this tab.
+      </p>
+    </PageLayout>
+  );
+};
+
+export default DeviceComplete;
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -33,7 +33,7 @@
         "typescript": "5.3.3"
       },
       "engines": {
-        "node": "18"
+        "node": "20"
       }
     },
     "node_modules/@aashutoshrathi/word-wrap": {


ui/pages/device_code.tsx
```diff
@@ -45,12 +49,21 @@ const DeviceCode: NextPage = () => {
...
 
+  if (isSuccess) {
+    return (
+      <PageLayout title="Sign in successful">
+        <p>
+          Your device has been successfully registered. You can close this tab.
+        </p>
+      </PageLayout>
+    );
+  }
```
nsklikas, line 65: The success needs to be in a new page.

The user will be redirected there after they have successfully logged in and given consent. Look at https://warthogs.atlassian.net/browse/IAM-735?focusedCommentId=456723 for a couple of examples.

edlerd: I am confused. This is rendered as a dedicated page as it was shown in the [design file](https://www.figma.com/file/eFR98XGAliMnPUWeBtWzKI/Login-UI?type=design&node-id=393-43679&mode=design&t=fmb6nbeCeP6zTMqz-4). See also the screen below. Or am I missing something?


image

nsklikas: Wait, now I am confused as well. I thought that this page is shown to the user when the user code is accepted (if it is accepted successfully we show this page).

Instead we need to show this to a user after they have given consent (the flow is: accept user_code -> login -> give consent).

So this needs to be on a new page (different file, different URL), without a condition.

edlerd: Oh I see, there is a step between entering the code and getting the success screen. :bulb: 

I'll change it to a dedicated page with dedicated url then, so it can be targetted directly.

ui/pages/device_complete.tsx
```diff
@@ -0,0 +1,15 @@
+import type { NextPage } from "next";
+import React from "react";
+import PageLayout from "../components/PageLayout";
+
+const DeviceComplete: NextPage = () => {
+  return (
+    <PageLayout title="Sign in successful">
+      <p>
+        Your device has been successfully registered. You can close this tab.
```
nsklikas, line 9: I don't like the word registered for this, I think we should use `connected`

edlerd: Yes, registered has a specific meaning in the identity domain, I agree connected is better and will make that change shortly.

nsklikas (dismissed):

nsklikas (approved):

PR #245, opened 2024-06-12
Add recovery and password settings flows

natalian98:
## Testing
```

make npm-build build

export PORT="4455"
export LOG_LEVEL="debug"

go run . serve
```
Go to localhost:4455/ui/reset_email:
image

The link to reset password is also added to the login screen:
image

Enter a valid email. I use `test@example.com` because that's the only imported identity in kratos.
Get the recovery code from mailslurper at localhost:4436.
image

Once a valid code is submitted, a kratos session is created. The recovery flow redirects to password change screen and starts a settings flow:
image
image

If you have an application that is a client of hydra, you will be automatically logged into it if opened in the same browser session as the completed recovery flow.
If you start a new browser session you can verify that the credential was updated.


natalian98:
There is a case that @edlerd is looking at:
If you are logged in and have a kratos session, you should be able to change your password without the need of going through recovery, but rn going directly to localhost:4455/ui/reset_password redirects to http://localhost:4455/ui/ which doesn't exist.

edlerd:
> There is a case that @edlerd is looking at:
> If you are logged in and have a kratos session, you should be able to change your password without the need of going through recovery, but rn going directly to localhost:4455/ui/reset_password redirects to http://localhost:4455/ui/ which doesn't exist.

This has been handled.

edlerd (approved):

pkg/kratos/service.go
```diff
@@ -120,6 +138,65 @@ func (s *Service) GetLoginFlow(ctx context.Context, id string, cookies []*http.C
...
+		UpdateRecoveryFlowBody(body).
+		Cookie(cookiesToString(cookies)).
+		Execute()
+
+	// We expect to get a 422 response from Kratos. The sdk forces us to
+	// make the request with an 'application/json' content-type, whereas Kratos
+	// expects the 'Content-Type' and 'Accept' to be 'application/x-www-form-urlencoded'.
+	// This is not a real error, as we still get the URL to which the user needs to be
+	// redirected to.
+	if err != nil && resp.StatusCode != 422 {
```
BarcoMasile, line 214: question: so this assumes that if `StatusCode` is actually 422 the `err` cannot be `nil`, meaning: this two conditions will happen together every time. Is this always the case?
If this is the case than we're good. If this is not the case than we would need an extra check on err

pkg/kratos/service.go
```diff
@@ -316,6 +393,42 @@ func (s *Service) ParseLoginFlowMethodBody(r *http.Request) (*kClient.UpdateLogi
...
+	err := parseBody(r.Body, &body)
+
+	if err != nil {
+		return nil, err
+	}
+	ret := kClient.UpdateRecoveryFlowWithCodeMethodAsUpdateRecoveryFlowBody(
+		body,
+	)
+
+	email := "test@example.com"
```
BarcoMasile: issue: I understand that this is the only existing user right now, but can we avoid hardcoding this and take this value from the UI form ? So this would work also for other account when there are some

docker/kratos/kratos.yml
```diff
@@ -1,9 +1,9 @@
-version: v1.0.0
+version: v1.1.0
 
 oauth2_provider:
     url: "http://hydra:4445"
 log:
-    level: debug
+    level: trace
```
BarcoMasile: question/possible issue: do we actually need `trace` level logging? This could be truly overwhelming in a production setting. Can't we just do `info`?

ui/pages/reset_password.tsx
```diff
@@ -1,12 +1,85 @@
...
+  }, [flowId, router, router.isReady, returnTo, flow]);
+  const handleSubmit = useCallback(
+    (values: UpdateSettingsFlowBody) => {
+      return kratos
+        .updateSettingsFlow({
+          flow: String(flow?.id),
+          updateSettingsFlowBody: values,
+        })
+        .then(async ({ data }) => {
+          console.log(data); // TODO: remove
```
BarcoMasile: this probably slipped

BarcoMasile (commented):
A few comments here and there, just need some clarifications but I think we're good to go

docker/kratos/kratos.yml
BarcoMasile: In next commits this gets changed back to `debug`, still my question stands

pkg/kratos/handlers.go
```diff
@@ -25,6 +25,10 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/kratos/self-service/login/browser", a.handleCreateFlow)
 	mux.Get("/api/kratos/self-service/login/flows", a.handleGetLoginFlow)
 	mux.Get("/api/kratos/self-service/errors", a.handleKratosError)
+	mux.Post("/api/kratos/self-service/recovery", a.handleUpdateRecoveryFlow)
+	mux.Get("/api/kratos/self-service/recovery/browser", a.handleCreateRecoveryFlow)
```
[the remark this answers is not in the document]
natalian98: It is by the kratos API docs: https://docs-omega-six.vercel.app/docs/kratos/reference/api#tag/frontend/operation/createBrowserRecoveryFlow

docker/kratos/kratos.yml
natalian98: Since this is a testing setup not intended to be used in production, I think `debug` is acceptable. In kratos charm the log level is configurable: https://github.com/canonical/kratos-operator/blob/7c6973a8fd9172f1fb2f22031f22f2883e234321/config.yaml#L48

pkg/kratos/service.go
natalian98: Yes, if the status code is 422, we always get a `browser location change required` error which is not a real one (as stated in the comment), it's just an indication that we need to change the redirect url

ui/pages/reset_password.tsx
natalian98: This is from an old commit, now removed

pkg/kratos/service.go
```diff
@@ -120,6 +155,92 @@ func (s *Service) GetLoginFlow(ctx context.Context, id string, cookies []*http.C
...
+		return nil, nil, err
+	}
+
+	// We trasform the kratos response to our own custom response here.
+	// The original kratos response contains an 'Error' field, which we remove
+	// because this is not a real error.
+	returnToResp := BrowserLocationChangeRequired{redirectResp.RedirectBrowserTo}
+
+	return &returnToResp, resp.Cookies(), nil
+}
```
nsklikas, line 242: I am not sure how I feel about this. Every time the user clicks a button ('Resend code` or `Submit`), this endpoint is called and it returns either `{}` or a url to redirect the user to.

The issue is that, every time right after this call the ui seems to call the get flow endpoint. Wouldn't it make sense if this endpoint returned the whole flow object to make 1 api call instead of 2?

Perhaps I am missing something

ui/pages/reset_email.tsx
```diff
@@ -1,30 +1,90 @@
...
+    }
+
+    // Otherwise we initialize it
+    kratos
+      .createBrowserRecoveryFlow({
+        returnTo: returnTo ? String(returnTo) : undefined,
+      })
+      .then(({ data }) => {
+        if (data.request_url !== undefined) {
+          window.location.href = data.request_url;
```
nsklikas, line 41: question: Why redirect there? This is the URL of the request that the backend made.

ui/pages/reset_password.tsx
```diff
@@ -1,31 +1,106 @@
...
+    }
+
+    // Otherwise we initialize it
+    kratos
+      .createBrowserSettingsFlow({
+        returnTo: returnTo ? String(returnTo) : undefined,
+      })
+      .then(({ data }) => {
+        if (data.request_url !== undefined) {
+          window.location.href = data.request_url;
```
nsklikas, line 45: question: Why redirect there? This is the URL of the request that the backend made.

nsklikas (commented):
Tried to play around a little and if a session already exists, going to http://localhost:4455/ui/reset_email will always redirect me to http://localhost:4455/ui/.

One other thing is that AFAICT sometimes the user-agent makes a call to the backend and then it makes a call (or redirects) to kratos. I didn't expect that the user would have to be redirected to kratos, I thought that all communications would go through the login-UI

pkg/kratos/handlers.go
```diff
@@ -193,6 +199,157 @@ func (a *API) handleKratosError(w http.ResponseWriter, r *http.Request) {
...
+	w.Write(resp)
+}
+
+func (a *API) handleCreateRecoveryFlow(w http.ResponseWriter, r *http.Request) {
+	returnTo, err := url.JoinPath(a.baseURL, "/ui/reset_email")
+	if err != nil {
+		a.logger.Fatal("Failed to construct returnTo URL: ", err)
+	}
+
+	flow, cookies, err := a.service.CreateBrowserRecoveryFlow(context.Background(), returnTo, r.Cookies())
```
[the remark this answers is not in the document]
natalian98: I might have misunderstood the [docs](https://github.com/ory/kratos-client-go/blob/c479681511b0b8193970d482c152ee59b1fc8b26/api_frontend.go#L1249) but I think it is required. The `return_to` in this case is the ui page for recovery

ui/pages/reset_email.tsx
natalian98: Without the redirect the flow id is not appended to the url

nsklikas: Doesn't this make the first call (the one done by the backend) useless? Can't the frontend just append the flow_id query param to the URL?

docker/kratos/kratos.yml
BarcoMasile: Oh ok, my bad! I didn't realize

BarcoMasile (approved):

edlerd:
In case no text or an invalid email is entered, the `POST api/kratos/self-service/recovery` responds with the text content `400 Bad Request`. As this is displayed to the user, it would be nice to have a better error message in this case.

pkg/kratos/service.go
natalian98: As discussed the redirect is needed to allow returning to settings after recovery is completed

pkg/kratos/service.go
```diff
@@ -159,6 +280,29 @@ func (s *Service) UpdateLoginFlow(
...
+	defer span.End()
+
+	settingsFlow, resp, err := s.kratos.FrontendApi().
+		UpdateSettingsFlow(ctx).
+		Flow(flow).
+		UpdateSettingsFlowBody(body).
+		Cookie(cookiesToString(cookies)).
+		Execute()
+
+	if err != nil && resp.StatusCode != 200 {
```
shipperizer, line 296:
```suggestion
	if err != nil && resp.StatusCode != http.StatusOK {
```

shipperizer, line 296:
```suggestion
	if err != nil || resp.StatusCode != 200 {
```

do we need the check on the `resp.StatusCode` at all?wouldn't err be enough?

shipperizer:
couple of small commit message nudges:

* try to bundle the tests in 4c7ee6912718c77142fe6d78c751be89d45936d8 with the code if u can
* first `wip` commit needs to become a `feat` or `fix`

pkg/kratos/service.go
natalian98: it may happen that kratos returns an "error" of info type on 200 response, in that case we want to ignore it

ui/pages/reset_email.tsx
```diff
@@ -1,30 +1,90 @@
...
+        .getRecoveryFlow({ id: String(flowId) })
+        .then((res) => setFlow(res.data))
+        .catch(handleFlowError(router, "recovery", setFlow));
+      return;
+    }
+
+    // Otherwise we initialize it
+    kratos
+      .createBrowserRecoveryFlow({
+        returnTo: returnTo ? String(returnTo) : undefined,
```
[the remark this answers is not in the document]
natalian98: It's required for redirecting to either `ui/reset_email?flow=xxx` or `/ui/reset_password` once recovery is completed

ui/pages/reset_email.tsx
```diff
@@ -1,30 +1,90 @@
...
+          }
+          if ("redirect_to" in data) {
+            window.location.href = data.redirect_to as string;
+            return;
+          }
+          if (flow?.return_to) {
+            window.location.href = flow.return_to;
+            return;
+          }
+          await router.push("/");
```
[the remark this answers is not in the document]
natalian98: This redirects to home page as fallback and it similarly handled in [login](https://github.com/canonical/identity-platform-login-ui/blob/65e63207deb03770e1b17a5278d585de1ea4b458/ui/pages/login.tsx#L88) flow

nsklikas:
> Tried to play around a little and if a session already exists, going to http://localhost:4455/ui/reset_email will always redirect me to http://localhost:4455/ui/.

This bug still exists

ui/pages/reset_password.tsx
```diff
@@ -1,31 +1,106 @@
...
+          return;
+        }
+        setFlow(data);
+      })
+      .catch(handleFlowError(router, "settings", setFlow))
+      .catch(async (err: AxiosError<string>) => {
+        if (err.response?.data.trim() === "Failed to create settings flow") {
+          setFlow(undefined);
+          await router.push("/login");
```
nsklikas, line 54:
```suggestion
          await router.push("./login");
```

When the UI implements the fix we discussed we can ditch the relative paths, but we need this for now

ui/pages/reset_password.tsx
```diff
@@ -1,31 +1,106 @@
...
+          flow: String(flow?.id),
+          updateSettingsFlowBody: {
+            csrf_token: (flow?.ui?.nodes[0].attributes as UiNodeInputAttributes)
+              .value as string,
+            method: "password",
+            password: password,
+          },
+        })
+        .then(async () => {
+          await router.push("/ui/reset_complete");
```
nsklikas, line 77:
```suggestion
          await router.push("./reset_complete");
```

ui/util/handleFlowError.ts
```diff
@@ -36,6 +36,11 @@ export const handleFlowError =
         // We need to re-authenticate to perform this action
         window.location.href = err.response.data.redirect_browser_to;
         return;
+      case "session_inactive":
+        // No session found, redirect the user to sign in
+        resetFlow(undefined);
+        await router.push("/login");
```
nsklikas, line 42:
```suggestion
        await router.push("./login");
```

nsklikas, line 42: I am not sure if this works right now. Currently the login logic is tied to hydra, if I try to go to `/ui/reset_password` right now without a session, I get redirected to this
image

The error on kratos is `The request was malformed or contained invalid parameters reason:the login_challenge parameter is present but empty`. The login-ui backend will need to change to handle this (probably on a different PR).

ui/components/NodeInputEmail.tsx
```diff
@@ -0,0 +1,30 @@
...
+      defaultValue={node.messages.map(({ text }) => text).join(" ")}
+      error={error}
+      onChange={(e) => void setValue(e.target.value)}
+      onKeyDown={(e) => {
+        if (e.key === "Enter") {
+          e.preventDefault();
+          e.stopPropagation();
+          void dispatchSubmit(e, "code");
+        }
+      }}
```
nsklikas, line 27: Would it make sense to add a check here for validating that it is an email?

Currently if the input is not an email, I get this
image

Kratos expects an email and because the input is not a valid email it returns an error that is not handled by the backend (the backend should handle this error when the error handling story is implemented)

pkg/kratos/service.go
```diff
@@ -43,6 +49,30 @@ type BrowserLocationChangeRequired struct {
...
+type UiNode []struct {
+	Messages Messages `json:"messages,omitempty"`
+}
+
+type UiErrorMessages struct {
+	Ui struct {
+		Messages Messages
+		Nodes    UiNode
+	} `json:"ui,omitempty"`
+}
```
nsklikas, line 74: I think this can be rewritten as
```suggestion
type UiErrorMessages struct {
    Ui kClient.UiContainer `json:"ui"`
}
```

ui/pages/reset_email.tsx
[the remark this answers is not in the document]
nsklikas: I think that we need to remove it from the `login` page as well, this page is not implemented

nsklikas (commented):
I think that once the frontend pathing is fixed (switch to relative paths) we can merge it and continue working on separate PRs (error handling, UX, etc).

ui/components/NodeInputEmail.tsx
edlerd: added frontend validation for the email input

ui/pages/reset_password.tsx
edlerd: replaced all the frontend paths with relative ones.

ui/util/handleFlowError.ts
natalian98: Issue opened: https://github.com/canonical/identity-platform-login-ui/issues/247

BarcoMasile (dismissed):

pkg/kratos/service.go
```diff
@@ -159,12 +289,50 @@ func (s *Service) UpdateLoginFlow(
...
-	errorMessages := new(kClient.LoginFlow)
+	errorMessages := new(UiErrorMessages)
 	body, _ := io.ReadAll(responseBody)
 	json.Unmarshal([]byte(body), &errorMessages)
 
 	errorCodes := errorMessages.Ui.Messages
+
+	// if no message was found, search through nodes
+	if len(errorCodes) == 0 {
+		nodes := errorMessages.Ui.Nodes
```
shipperizer, line 324: we don't perform any validation on the `errorMessages.Ui` so to play it on the safe side we should proabbly use the getter method `GetNodes` same got `node.Messages`

pkg/kratos/handlers.go
```diff
@@ -193,6 +199,159 @@ func (a *API) handleKratosError(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	resp, err := flow.MarshalJSON()
+	if err != nil {
+		a.logger.Errorf("Error when marshalling json: %v\n", err)
+		http.Error(w, "Failed to marshal json", http.StatusInternalServerError)
+		return
+	}
+	setCookies(w, cookies)
```
shipperizer, line 350: you might want to drop https://github.com/canonical/identity-platform-login-ui/pull/245/files#diff-12151c8a9259cc64c5b82a9a51ed3eb6367a4850988ecd45cb323b4e086c708cR372

shipperizer (commented):
not sure what is missing on the frontend side but if you guys @natalian98 @nsklikas @edlerd are happy with this as a start i'm happy to approve

ui/pages/reset_email.tsx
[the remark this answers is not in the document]
edlerd: Changed it here and in login to redirect to the error page

pkg/kratos/handlers.go
natalian98: This handler is still used in the login flow (see [example](https://github.com/canonical/identity-platform-login-ui/blob/65e63207deb03770e1b17a5278d585de1ea4b458/pkg/kratos/service.go#L52)). Can I refactor it in another PR since I'm addressing recovery and settings in this one?
Some other comments you made or `ioutil` deprecation are also relevant for the login flow

shipperizer: missed that, yep that's fair, goog to go then

shipperizer:
@natalian98 do a rebase and then it's mergeable

nsklikas (approved):

PR #249, opened 2024-06-28
Support multi factor authentication

natalian98:
## Testing
```
make npm-build build

export PORT="4455"
export LOG_LEVEL="debug"

go run . serve

```
### Settings
Go to localhost:4455/ui/reset_email and go through the recovery flow (test@example.com). You will first be redirected to change the password:
image
Then to set up MFA:
image
After linking the totp app:
image

When MFA is set up:
- the authenticator app can be unlinked:
image
- on subsequent recovery flows, user will be required to provide the second factor before changing settings. In future PRs I will add lookup secrets to have a backup method in case the access to the authenticator app was lost.

### Login
When MFA is configured, you will be required to provide it when signing in:
image

When logged in, you can access the settings at localhost:4455/ui/reset_password or localhost:4455/ui/setup_secure.


pkg/kratos/service.go
```diff
@@ -504,16 +517,53 @@ func (s *Service) ParseRecoveryFlowMethodBody(r *http.Request) (*kClient.UpdateR
 }
 
 func (s *Service) ParseSettingsFlowMethodBody(r *http.Request) (*kClient.UpdateSettingsFlowBody, error) {
-	body := new(kClient.UpdateSettingsFlowWithPasswordMethod)
+	type MethodOnly struct {
+		Method string `json:"method"`
+	}
```
BarcoMasile, line 539: small suggestion: this struct is only used here, we could define it once outside of the function, and make it start with a lowercase letter so it doesn't end up getting exported

ui/pages/setup_secure.tsx
```diff
@@ -1,19 +1,92 @@
...
+
+    // Otherwise we initialize it
+    kratos
+      .createBrowserSettingsFlow({
+        returnTo: returnTo ? String(returnTo) : undefined,
+      })
+      .then(({ data }) => {
+        if (data.request_url !== undefined) {
+          window.location.href =
+            "http://localhost:4455/ui/setup_secure" + "?flow=" + data.id;
```
BarcoMasile: issue: host and port should be externalized piece of info
this could easily break

pkg/kratos/service.go
```diff
@@ -55,12 +55,12 @@ type UiErrorMessages struct {
...
 
 func (s *Service) CheckSession(ctx context.Context, cookies []*http.Cookie) (*kClient.Session, []*http.Cookie, error) {
-	ctx, span := s.tracer.Start(ctx, "kratos.FrontendApi.ToSession")
+	ctx, span := s.tracer.Start(ctx, "kratos.Service.ToSession")
 	defer span.End()
 
 	session, resp, err := s.kratos.FrontendApi().
 		ToSession(ctx).
-		Cookie(cookiesToString(cookies)).
+		Cookie(httpHelpers.CookiesToString(cookies)).
```
BarcoMasile, line 63: question: does this `CookiesToString` function take care also of other attributes for the cookies like path, domain, expires etc.., or is it on the sdk to take care of that?

natalian98: It doesn't - it just gets the cookies names and values (see [here](https://github.com/canonical/identity-platform-login-ui/blob/3965a2f2c05aa49d0f8d446190ad07a8904fc70c/internal/misc/http/helpers.go#L17), the rest is on sdk side

pkg/kratos/service.go
natalian98: I think this field must be exported, because it has a json tag. If `method` is not capitalized then json package won't see its value

ui/pages/setup_secure.tsx
edlerd: You are right, this should be a relative redirect. Pushed a commit to fix it.

BarcoMasile (dismissed):

pkg/kratos/service.go
```diff
@@ -180,7 +185,7 @@ func (s *Service) GetRecoveryFlow(ctx context.Context, id string, cookies []*htt
 	return flow, resp.Cookies(), nil
 }
 
-func (s *Service) GetSettingsFlow(ctx context.Context, id string, cookies []*http.Cookie) (*kClient.SettingsFlow, []*http.Cookie, error) {
+func (s *Service) GetSettingsFlow(ctx context.Context, id string, cookies []*http.Cookie) (*kClient.SettingsFlow, []*http.Cookie, *BrowserLocationChangeRequired, error) {
```
shipperizer, line 188: this methods seems to be doing too much

we have 4 returning parameters which in a lot of cases are `nil` 

would be worth refactoring that and in case duplicate logic or move some of the discriminating logic one layer up into the handler

nsklikas:
After I input the new password, I get redirected to http://localhost:4455/setup_secure
image

Does this only happen to me? I use docker to run external components and run the login-ui app using `go run`.

natalian98:
> After I input the new password, I get redirected to http://localhost:4455/setup_secure image
> 
> Does this only happen to me? I use docker to run external components and run the login-ui app using `go run`.

You are right, [this](https://github.com/canonical/identity-platform-login-ui/pull/249/commits/274558f060d4b61bab926def566eeb71a0fdf72a#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9R54) change broke the redirect.
Adding `/ui` to the path: `./ui/setup_secure?flow=${data.id}${pwParam}` fixes it, @edlerd could you advise why relative paths don't work here?

edlerd:
> You are right, [this](https://github.com/canonical/identity-platform-login-ui/pull/249/commits/274558f060d4b61bab926def566eeb71a0fdf72a#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9R54) change broke the redirect.
> Adding `/ui` to the path: `./ui/setup_secure?flow=${data.id}${pwParam}` fixes it, @edlerd could you advise why relative paths don't work here?

I changed all redirects to not rely on the Next.js router. The next router ignores the current path, hence switching to the vanilla js solution. When I tested it, all redirects work as expected.

nsklikas (commented):
Looks like there are still a couple of issues with this. 

After I instantiate the authenticator when I try to do the reset flow, the reset password page is momentarily shown and then I immediately get to the login page where I have to do input the code form the authenticator. BUT after I enter the code I get this:
image
The  error on kratos is 
```
error=map[message:I[#/] S[] could not find a strategy to login with stack_trace:
```

Also after I do the reset flow, if I try to redo the reset flow by going again to `reset_email` I will get an error. I thought that we had handled this in the previous PR

pkg/kratos/handlers.go
```diff
@@ -72,7 +70,7 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 
 	returnTo, err := url.JoinPath(a.baseURL, "/ui/login")
 	if err != nil {
-		a.logger.Fatal("Failed to construct returnTo URL: ", err)
+		a.logger.Error("Failed to construct returnTo URL: ", err)
```
shipperizer, line 73:
```suggestion
		a.logger.Errorf("Failed to construct returnTo URL: %v", err)
```

nsklikas:
Had a look at this https://github.com/canonical/identity-platform-login-ui/pull/249#pullrequestreview-2153619712, looks like there is something wrong with the state handling in the flow object (https://github.com/canonical/identity-platform-login-ui/blob/main/ui/components/Flow.tsx#L111). This is used to populate the method field in the request. For some reason some times this is `password` and some times it is `totp` (it should be `totp`). I am not sure how this works, but maybe we should explicitly specify the flow method in the ui when making the call (https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/login.tsx#L77)

For some reason natalia was not able to reproduce this (maybe different node versions?), but it happens consistently to me.

What do you think @edlerd @natalian98? We can have a chat tomorrow as I think it will be easier to show you, rather than describe it via text.

pkg/kratos/handlers.go
```diff
@@ -280,19 +278,29 @@ func (a *API) handleCreateRecoveryFlow(w http.ResponseWriter, r *http.Request) {
...
-	resp, err := flow.MarshalJSON()
+	resp, err := json.Marshal(flow)
 	if err != nil {
 		a.logger.Errorf("Error when marshalling json: %v\n", err)
 		http.Error(w, "Failed to parse settings flow", http.StatusInternalServerError)
 		return
 	}
+
+	if redirect != nil {
+		resp = []byte(*redirect.RedirectTo)
```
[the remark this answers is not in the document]
natalian98: If it's a json it doesn't redirect:
image
I think that's because errors are handled differently in `reset_password` than in login: https://github.com/canonical/identity-platform-login-ui/pull/249/files#diff-8644072735eaac8081eaf42339bad3844a0caa4c0fe45d89a987e9d10e039284R37
In [login](https://github.com/canonical/identity-platform-login-ui/blob/3965a2f2c05aa49d0f8d446190ad07a8904fc70c/ui/pages/login.tsx#L54), the redirect url is read from a `FlowResponse` type.

I'm fine to return the json here, but can we parse the error responses in the same way @edlerd?

pkg/kratos/handlers.go
```diff
@@ -278,30 +278,42 @@ func (a *API) handleCreateRecoveryFlow(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
 
-	if redirect != nil {
-		resp = []byte(*redirect.RedirectTo)
+		// We trasform the kratos response to our own custom response here.
+		// The original kratos response contains an 'Error' field, which we remove
+		// because this is not a real error.
+		returnToResp := BrowserLocationChangeRequired{redirectResp.RedirectBrowserTo}
+		resp := []byte(*returnToResp.RedirectTo)
```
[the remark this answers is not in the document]
natalian98: Yes I will change that, but the frontend error parsing needs to be adjusted, see [this](https://github.com/canonical/identity-platform-login-ui/pull/249#discussion_r1664053221) comment

pkg/kratos/handlers.go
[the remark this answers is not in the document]
edlerd: Yes, we can change this format. The current implementation checks if the response body is a valid url and if that is paired with a 403 status, we redirect to that url. changing this response to a json would mean we also have to change the error handling. Let me know if you change the response body structure and I can also change the frontend error handling.

edlerd:
> Had a look at this [#249 (review)](https://github.com/canonical/identity-platform-login-ui/pull/249#pullrequestreview-2153619712), looks like there is something wrong with the state handling in the flow object (https://github.com/canonical/identity-platform-login-ui/blob/main/ui/components/Flow.tsx#L111). This is used to populate the method field in the request. For some reason some times this is `password` and some times it is `totp` (it should be `totp`). I am not sure how this works, but maybe we should explicitly specify the flow method in the ui when making the call (https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/login.tsx#L77)
> 
> For some reason natalia was not able to reproduce this (maybe different node versions?), but it happens consistently to me.
> 
> What do you think @edlerd @natalian98? We can have a chat tomorrow as I think it will be easier to show you, rather than describe it via text.

Pushing a fix for fixating the method based on the fields that are displayed here. This should fix the error you are seeing @nsklikas

nsklikas:
Now when I try to reset the password, with a user that has configured mfa, I input the recovery code, I don't get shown the otp page, I get shown the reset password page and the get flows call fails with:
```
{
  "error": {
    "code": 403,
    "details": {
      "redirect_browser_to": "http://localhost:4433/self-service/login/browser?aal=aal2u0026return_to=http%3A%2F%2Flocalhost%3A4455%2Fui%2Freset_password%3Fflow%3D8b22e3fc-7258-4dea-91b4-652617dc1858"
    },
    "id": "session_aal2_required",
    "message": "Session does not fulfill the requested Authenticator Assurance Level",
    "reason": "An active session was found but it does not fulfill the requested Authenticator Assurance Level. Please verify yourself with a second factor to resolve this issue.",
    "status": "Forbidden"
  },
  "redirect_browser_to": "http://localhost:4433/self-service/login/browser?aal=aal2u0026return_to=http%3A%2F%2Flocalhost%3A4455%2Fui%2Freset_password%3Fflow%3D8b22e3fc-7258-4dea-91b4-652617dc1858"
}
```

When I try to login I see only this:
image

ui/pages/reset_password.tsx
```diff
@@ -31,7 +32,15 @@ const ResetPassword: NextPage = () => {
...
-        .catch(handleFlowError(router, "settings", setFlow));
+        .catch(handleFlowError("settings", setFlow))
+        .catch((err: AxiosError<string>) => {
+          if (err.response?.status === 403 && isValidUrl(err.response.data)) {
+            window.location.href = err.response.data;
+            return;
+          }
+
+          return Promise.reject(err);
+        });
```
nsklikas, line 43: minor: When I try login to reset the password of a user who has enabled mfa, I momentarily see the reset_password page and then I get redirected to the authentication code page and then I log in and reset the password. My question is whether we could move this earlier so that the reset_password page does not load at all if the user needs to be redirected?

nsklikas:
> Pushing a fix for fixating the method based on the fields that are displayed here. This should fix the error you are seeing @nsklikas

@edlerd looks good, can't reproduce it now

ui/pages/reset_password.tsx
edlerd: As this request to fetch the flow happens only after the initial render we need to show something. That could be a loading screen, but I am not sure that will be a better experience -- seeing the loading screen / or a blank screen momentarily.

pkg/kratos/handlers.go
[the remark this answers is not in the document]
natalian98: Response body changed, it is now a json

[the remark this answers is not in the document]
edlerd: Changed the frontend to handle the new JSON response body instead of the plain URL in the body.

ui/pages/reset_password.tsx
edlerd: Changed this to show a blank page as long as the flow is loading. So we don't have a flicker of showing the "reset password" form in between landing and the redirect.

nsklikas (approved):

pkg/kratos/handlers.go
```diff
@@ -280,20 +283,39 @@ func (a *API) handleCreateRecoveryFlow(w http.ResponseWriter, r *http.Request) {
...
+		redirectResp := new(ErrorBrowserLocationChangeRequired)
+		err = unmarshalByteJson(response.Body, redirectResp)
+		if err != nil {
+			a.logger.Errorf("Failed to unmarshal JSON: %s", err)
+			return
+		}
+
+		if redirectResp.HasRedirectTo() {
+			a.logger.Errorf("Error when getting settings flow: %v\n", response)
+			w.WriteHeader(http.StatusInternalServerError)
```
nsklikas, line 304: Why is this a `500`? AFAICT the request was processed fine, but the user session is not ready. Shouldn't it be 4XX?

pkg/kratos/handlers.go
```diff
@@ -328,26 +350,43 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
+		err = unmarshalByteJson(response.Body, redirectResp)
+		if err != nil {
+			a.logger.Errorf("Failed to unmarshal JSON: %s", err)
+			return
+		}
+
+		if redirectResp.HasRedirectTo() {
+			a.logger.Errorf("Error when creating settings flow: %v\n", response)
+			w.WriteHeader(http.StatusInternalServerError)
+			_ = json.NewEncoder(w).Encode(redirectResp)
```
nsklikas, line 378: Why is this a `500`? AFAICT the request was processed fine, but the user session is not ready. Shouldn't it be 4XX?

pkg/kratos/handlers.go
```diff
@@ -328,26 +350,43 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
-	flow, cookies, err := a.service.CreateBrowserSettingsFlow(context.Background(), returnTo, r.Cookies())
+	flow, response, err := a.service.CreateBrowserSettingsFlow(context.Background(), returnTo, r.Cookies())
 	if err != nil {
 		a.logger.Errorf("Failed to create settings flow: %v", err)
 		http.Error(w, "Failed to create settings flow", http.StatusInternalServerError)
 		return
 	}
 
+	// If aal1, redirect to complete second factor auth
+	if response.StatusCode == http.StatusForbidden {
```
nsklikas, line 367: IMO This is logic does not belong to the handlers, we should move it to the service layer.

pkg/kratos/service.go
```diff
@@ -129,7 +143,7 @@ func (s *Service) CreateBrowserRecoveryFlow(ctx context.Context, returnTo string
 	return flow, resp.Cookies(), nil
 }
 
-func (s *Service) CreateBrowserSettingsFlow(ctx context.Context, returnTo string, cookies []*http.Cookie) (*kClient.SettingsFlow, []*http.Cookie, error) {
+func (s *Service) CreateBrowserSettingsFlow(ctx context.Context, returnTo string, cookies []*http.Cookie) (*kClient.SettingsFlow, *http.Response, error) {
```
nsklikas, line 146: I don't like having to change the service function's signature. If we end up changing it, (imho) we will need to do it for all the service functions.

pkg/kratos/handlers.go
shipperizer: I'm leaning towards Nikos position as well, we are supposed to wrap kratos SDK and hide the response behind the service layer

pkg/kratos/service.go
```diff
@@ -189,17 +206,20 @@ func (s *Service) GetSettingsFlow(ctx context.Context, id string, cookies []*htt
...
 	}
 
-	return flow, resp.Cookies(), nil
+	return flow, resp, nil
 }
 
 func (s *Service) UpdateRecoveryFlow(
 	ctx context.Context, flow string, body kClient.UpdateRecoveryFlowBody, cookies []*http.Cookie,
-) (*BrowserLocationChangeRequired, []*http.Cookie, error) {
+) (*ErrorBrowserLocationChangeRequired, []*http.Cookie, error) {
```
shipperizer, line 222: feels weird that the main object we return is called Error<something>

pkg/kratos/service.go
shipperizer: I agree with Nikos, the response from.kratos.should be dealt with at the service layer level

BarcoMasile (dismissed):

shipperizer (dismissed):
code looks good to go

nsklikas (approved):
lgtm, minor improvement comments

nsklikas (dismissed):

pkg/kratos/handlers.go
```diff
@@ -328,28 +347,43 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
+	// If aal1, redirect to complete second factor auth
+	if response.HasRedirectTo() {
+		a.logger.Errorf("Failed to create settings flow due to insufficient aal: %v\n", response)
+		w.WriteHeader(http.StatusForbidden)
+		_ = json.NewEncoder(w).Encode(
+			ErrorBrowserLocationChangeRequired{
+				Error:             response.Error,
+				RedirectBrowserTo: response.RedirectTo,
+			},
+		)
```
shipperizer, line 384:
```suggestion
	// If aal1, redirect to complete second factor auth
	if response != nil && response.HasRedirectTo() {
		a.logger.Errorf("Failed to create settings flow due to insufficient aal: %v\n", response)
		w.WriteHeader(http.StatusForbidden)
		_ = json.NewEncoder(w).Encode(
			ErrorBrowserLocationChangeRequired{
				Error:             response.Error,
				RedirectBrowserTo: response.RedirectTo,
			},
		)
	
	if flow == nil {
	   <return internal server error>
	}
	
	resp, err := flow.MarshalJSON()
	if err != nil {
		a.logger.Errorf("Error when marshalling json: %v\n", err)
		http.Error(w, "Failed to marshal json", http.StatusInternalServerError)
		return
	}
	w.WriteHeader(http.StatusOK)
	w.Write(resp)
	return
}
```

shipperizer (approved):

PR #254, opened 2024-07-12
ui: set oidc method on login if a provider is present in the values

edlerd:


natalian98 (approved):
Thanks for the PR @edlerd, that fixes the issue. For future webauthn support we may need to determine the flow method based on different parameters (I suppose the nodes `group` would work).

PR #255, opened 2024-07-23
Add passwordless webauthn support

natalian98:
### Testing
```
make npm-build build

export PORT="4455"
export LOG_LEVEL="debug"

go run . serve

```

```
docker exec <hydra-container> \
  hydra create client \
    --endpoint http://127.0.0.1:4445 \
    --name grafana \
    --grant-type authorization_code,refresh_token \
    --response-type code,id_token \
    --format json \
    --scope openid,offline_access,email \
    --redirect-uri http://localhost:2345/login/generic_oauth

docker run -d --name=grafana -p 2345:2345 --network identity-platform-login-ui_intranet \
-e "GF_SERVER_HTTP_PORT=2345" \
-e "GF_AUTH_GENERIC_OAUTH_ENABLED=true" \
-e "GF_AUTH_GENERIC_OAUTH_AUTH_ALLOWED_DOMAINS=hydra,localhost" \
-e "GF_AUTH_GENERIC_OAUTH_NAME=Identity Platform" \
-e "GF_AUTH_GENERIC_OAUTH_CLIENT_ID=<client-id>" \
-e "GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET=<client-secret>" \
-e "GF_AUTH_GENERIC_OAUTH_SCOPES=openid offline_access email" \
-e "GF_AUTH_GENERIC_OAUTH_AUTH_URL=http://localhost:4444/oauth2/auth" \
-e "GF_AUTH_GENERIC_OAUTH_TOKEN_URL=http://hydra:4444/oauth2/token" \
-e "GF_AUTH_GENERIC_OAUTH_API_URL=http://hydra:4444/userinfo" \
grafana/grafana
```
Go to localhost:2345 and sign in (`test@example.com / test`), then connect the account with a passkey at localhost:4455/ui/setup_passkey:
image

Open a new browser session and sign in with passkey:
image

image

Use `test@example.com` account to sign in:
image

image

image

If request is cancelled:
image

### DIFF
--- a/ui/components/NodeInputButton.tsx
+++ b/ui/components/NodeInputButton.tsx
@@ -3,6 +3,20 @@ import { Button } from "@canonical/react-components";
 import React, { FC, FormEvent } from "react";
 import { NodeInputProps } from "./helpers";
 
+const getWebAuthnPayload = (evalCode: string): unknown => {
+  const tmp = evalCode
+    .replace("window.__oryWebAuthnRegistration(", "")
+    .replace("window.__oryWebAuthnLogin(", "");
+  const raw = tmp.substring(0, tmp.length - 1);
+  return JSON.parse(raw) as unknown;
+};
+
+interface WebauthnWindow extends Window {
+  __oryWebAuthnRegistration: (a: unknown) => void;
+  __oryWebAuthnLogin: (a: unknown) => void;
+  __oryStartedLogin: boolean;
+}
+
 export const NodeInputButton: FC<NodeInputProps> = ({
   node,
   attributes,
@@ -10,13 +24,96 @@ export const NodeInputButton: FC<NodeInputProps> = ({
   disabled,
   dispatchSubmit,
 }) => {
+  const onClick = attributes.onclick ?? "";
+
+  const startWebAuthnLogin = () => {
+    const webauthnWindow = window as unknown as WebauthnWindow;
+    const loginParams = getWebAuthnPayload(onClick);
+
+    if (webauthnWindow.__oryStartedLogin) {
+      return;
+    }
+    webauthnWindow.__oryStartedLogin = true;
+
+    // retry login until the webauthn.js script is finished loading
+    // and the __oryWebAuthnLogin function is available
+    const triggerWebauthnLogin = (count: number) =>
+      setTimeout(() => {
+        if ("__oryWebAuthnLogin" in webauthnWindow) {
+          webauthnWindow.__oryWebAuthnLogin(loginParams);
+        } else if (count < 100) {
+          triggerWebauthnLogin(count + 1);
+        }
+      }, 100);
+
+    triggerWebauthnLogin(0);
+  };
+
+  // automatically start the webauthn login
+  if (
+    onClick?.startsWith("window.__oryWebAuthnLogin(") &&
+    getNodeLabel(node) === "Continue"
+  ) {
+    startWebAuthnLogin();
+  }
+
   const handleClick = (e: MouseEvent | FormEvent) => {
+    // webauthn has a custom handler in webauthn.js to be called
+    // preventing the call of eval with below code
+    if (onClick?.startsWith("window.__oryWebAuthnRegistration(")) {
+      e.stopPropagation();
+      e.preventDefault();
+
+      const nameSelector = '*[name="webauthn_register_displayname"]';
+      const name = document.querySelector(nameSelector) as HTMLInputElement;
+
+      if (!name?.value) {
+        name.classList.add("is-error");
+        name.setAttribute("aria-invalid", "true");
+        name.setAttribute("aria-errormessage", "Required field");
+
+        if (!name.nextElementSibling) {
+          const warning = document.createElement("p");
+          warning.className = "p-form-validation__message";
+          warning.textContent = "Required field";
+          name.after(warning);
+        }
+
+        const groupSelector = ".p-form-validation";
+        const group = document.querySelectorAll(groupSelector);
+        group.forEach((e) => e.classList.add("is-error"));
+
+        return;
+      }
+
+      const webauthnWindow = window as unknown as WebauthnWindow;
+      const registrationParams = getWebAuthnPayload(onClick);
+      webauthnWindow.__oryWebAuthnRegistration(registrationParams);
+
+      return;
+    }
+
+    if (onClick?.startsWith("window.__oryWebAuthnLogin(")) {
+      e.stopPropagation();
+      e.preventDefault();
+
+      const webauthnWindow = window as unknown as WebauthnWindow;
+      const loginParams = getWebAuthnPayload(onClick);
+      webauthnWindow.__oryWebAuthnLogin(loginParams);
+
+      return;
+    }
+
     void setValue(attributes.value as string).then(() => dispatchSubmit(e));
   };
 
   return (
     <>
-      <Button onClick={handleClick} disabled={attributes.disabled || disabled}>
+      <Button
+        onClick={handleClick}
+        disabled={attributes.disabled || disabled}
+        name={attributes.name}
+      >
         {getNodeLabel(node)}
       </Button>
     </>
--- a/ui/components/NodeInputUrl.tsx
+++ b/ui/components/NodeInputUrl.tsx
@@ -1,0 +1,33 @@
+import { Button } from "@canonical/react-components";
+import React, { FC } from "react";
+import { NodeInputProps } from "./helpers";
+
+export const NodeInputUrl: FC<NodeInputProps> = ({ node }) => {
+  return (
+    <Button
+      onClick={(e) => {
+        e.preventDefault();
+        e.stopPropagation();
+
+        // read email from input field
+        const email =
+          (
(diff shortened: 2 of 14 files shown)


ui/pages/setup_passkey.tsx
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        return Promise.reject(err);
+      });
+  }, [flowId, router, router.isReady, returnTo, flow]);
+
+  const handleSubmit = useCallback(
+    (values: UpdateSettingsFlowBody) => {
+      const methodValues = values as UpdateSettingsFlowWithWebAuthnMethod;
+      console.log(values);
```
BarcoMasile: suggestion: this should be removed

ui/pages/setup_passkey.tsx
```diff
@@ -0,0 +1,129 @@
...
+      }),
+    },
+  } as SettingsFlow;
+
+  // TODO: The webauthn script should be launched on submit
+  return (
+    <PageLayout title="Set up a passkey login method">
+      <Head>
+        <script
+          src="http://localhost:4433/.well-known/ory/webauthn.js"
```
BarcoMasile: issue: this needs to be externalized, we can't rely on hardcoded host and ports

ui/pages/setup_passkey.tsx
```diff
@@ -62,40 +57,11 @@ const SetupPasskey: NextPage = () => {
...
-          }
-          await router.push("./error");
-        })
-        .catch(handleFlowError("settings", setFlow));
-    },
-    [flow, router],
-  );
+  const handleSubmit = (values: UpdateSettingsFlowBody) => {
+    // this is handled by the webauthn script
+    console.log(values);
```
BarcoMasile: suggestion: this should be removed

ui/components/NodeInputButton.tsx
```diff
@@ -11,7 +11,14 @@ export const NodeInputButton: FC<NodeInputProps> = ({
   dispatchSubmit,
 }) => {
   const handleClick = (e: MouseEvent | FormEvent) => {
-    void setValue(attributes.value as string).then(() => dispatchSubmit(e));
+    if (attributes.onclick) {
+      // handle passkeys that bring their own trigger
+      e.stopPropagation();
+      e.preventDefault();
+      eval(attributes.onclick);
```
BarcoMasile: question: isn't there any other way to do this than with `eval`? Can we just invoke cast `attributes.onclick` to the right signature function and invoke it? @edlerd

ui/components/NodeInputButton.tsx
```diff
@@ -11,14 +21,22 @@ export const NodeInputButton: FC<NodeInputProps> = ({
...
+    if (onClick?.startsWith("window.__oryWebAuthnRegistration(")) {
       e.stopPropagation();
       e.preventDefault();
-      eval(attributes.onclick);
-    } else {
-      void setValue(attributes.value as string).then(() => dispatchSubmit(e));
+
+      const webauthnWindow = window as unknown as WebauthnWindow;
+      const registrationParams = getWebAuthnRegistrationPayload(onClick);
+      webauthnWindow.__oryWebAuthnRegistration(registrationParams);
```
BarcoMasile: question: I see no guard for this cast. Are we sure this will always succeed in this context?

ui/components/NodeInputButton.tsx
```diff
@@ -3,6 +3,16 @@ import { Button } from "@canonical/react-components";
 import React, { FC, FormEvent } from "react";
 import { NodeInputProps } from "./helpers";
 
+const getWebAuthnRegistrationPayload = (evalCode: string): unknown => {
+  const tmp = evalCode.replace("window.__oryWebAuthnRegistration(", "");
+  const raw = tmp.substring(0, tmp.length - 1);
+  return JSON.parse(raw) as unknown;
+};
```
BarcoMasile: question: sorry but I must ask: don't we have any other way to do this?

ui/components/NodeInputButton.tsx
```diff
@@ -20,17 +24,45 @@ export const NodeInputButton: FC<NodeInputProps> = ({
...
 
-    // webauthn registration has a custom handler in webauthn.js to be called
+    // retry login until the webauthn.js script is finished loading
+    // and the __oryWebAuthnLogin function is available
+    const triggerWebauthnLogin = (count: number) =>
+      setTimeout(() => {
+        if ("__oryWebAuthnLogin" in webauthnWindow) {
+          webauthnWindow.__oryWebAuthnLogin(loginParams);
+        } else if (count < 100) {
+          triggerWebauthnLogin(count + 1);
```
BarcoMasile, line 45: question(possible-issue): this waits for the login func to be available, but my question is, since not all browser seem to support tail call optimization (ECMAScript 6), are we sure we don't risk any stack issues in case the webauthn.js script is slow to load or doesn't load at all (like hydra is not available or something) ?

ui/components/NodeInputUrl.tsx
```diff
@@ -0,0 +1,33 @@
...
+          (
+            document.querySelector(
+              'input[name="identifier"]',
+            ) as HTMLInputElement
+          )?.value || "";
+
+        // redirect to page with same url as current page, append webauthn query param
+        window.location.href = `${window.location.href}&webauthn=true&email=${email}`;
+      }}
+      className="oidc-login-button"
```
BarcoMasile, line 23: suggestion: 2 things:
- this `onClick` can benefit from a `useCallback`
- since this seems to have webauthn-related logic, I suggest we name this something less generic, so it's clear this is intended only for webauthn

BarcoMasile (changes requested):
Some question and minor comments, the required change is to avoid hardcoded host and port reference for Hydra

BarcoMasile:
I see that some of my comments already appear as `Outdated`. This is because common changes were not squashed in the UI code. We should always squash when possible so we don't end up reviewing "stale" code
Can you please fix that? @edlerd

ui/pages/setup_passkey.tsx
edlerd: has been removed.

ui/pages/setup_passkey.tsx
edlerd: was removed

ui/components/NodeInputButton.tsx
edlerd: we don't do the eval anymore

ui/components/NodeInputButton.tsx
edlerd: If we want to avoid the eval, we need to have a custom parser. If you have an idea how to improve this?

ui/components/NodeInputButton.tsx
edlerd: We can reduce the max nesting level, but I think 100 should be safe even with legacy browsers.

ui/components/NodeInputUrl.tsx
edlerd: Having a url type component as a Node in the Flow is a generic concept. I agree, this redirect function is now very tailored to the webauthn use case and its usage is not nice. Do you have a better idea how to implement this passkey button that appears as oidc option, that also reads the email value?

edlerd:
> I see that some of my comments already appear as `Outdated`. This is because common changes were not squashed in the UI code. We should always squash when possible so we don't end up reviewing "stale" code Can you please fix that? @edlerd

Squashed all my previous commits into a single one.

@natalian98 please resign your recent two commits as they are now based on a new root commit from mine and the signing became invalid.

ui/components/NodeInputButton.tsx
edlerd: If the webauhtn.js script is loaded, the functions will be available. As this is on click, we can be optimistic the script is finished loading at the time of execution. The network request for the script might have failed, leading to an error here. If we guard it with a check if the function on the window object is available and the check is false, I fail to have a better idea that showing a failure -- which should be the same current behaviour.

ui/components/NodeInputButton.tsx
BarcoMasile: Custom parser would be out of the discussion, so I'd say we don't have another way of doing this

ui/components/NodeInputUrl.tsx
BarcoMasile: I was not commenting the logic but the naming of the component.
It could be renamed to `WebAuthnNodeInputUrl`.

My suggestion for the use of the `useCallback` still stands

ui/components/NodeInputButton.tsx
BarcoMasile: That's ok. What kind of failure would the user see/get in case the function is not available?

BarcoMasile (dismissed):

ui/components/NodeInputUrl.tsx
edlerd: Added the callback and refactored this component so it appears as more generic.

shipperizer (approved):

BarcoMasile (approved):

PR #265, opened 2024-08-07
IAM-976 Implement MFA toggle and enforcement with redirection to `setup_secure`

BarcoMasile:
## Description
This PR introduces the MFA_ENABLED env var to enable MFA for authentication with Kratos as a local provider.
Even if MFA_ENABLED=true the login process through external provider is not affected.
Right now to enforce mfa the app just redirects to `/ui/setup_secure` (also considering any context path the app is configure with through the full URL env var called `BASE_URL`).
Next step will be to align with the UI team to agree on a specific error to show the user with a call to action to go and setup MFA instead of directly redirecting the user.

NB: the check on wether the user has TOTP enabled or not is done through Kratos admin identity APIs.
Kratos doesn't allow to enforce a specific AAL (Authenticator Assurance Level), but only allows to enforce the highest availalbe for the specific user. So if a user doesn't have TOTP setup (AAL2), he/she will be allowed to login with only username and password (for local provider), which is AAL1. That's why we need to make it so the user gets "forced" to setup MFA.
And then he/she can **try login again**.

## New environment variables
- `MFA_ENALBED`
- `KRATOS_ADMIN_URL`

## Additional changes
- some refactoring and fixes on the use of golang `context.Context` from http requests
- streamlined both logging and comments to ease reading of the code
- removed kratos logic from hydra client and reused existing kratos service


nsklikas (commented):
Not many comments from a code perspective, but I have a question:
Why are we checking the kratos session on the consent endpoint? I would think that kratos should be involved only in the login flow.

pkg/extra/service.go
```diff
@@ -2,45 +2,24 @@ package extra
...
-	).Execute()
-
-	if err != nil {
-		// TODO @shipperizer we shouldn't be logging this
-		s.logger.Debugf("full HTTP response: %v", r)
-
-		return nil, err
-	}
-	return session, nil
-}
```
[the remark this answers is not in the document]
BarcoMasile: So I removed this because it was copy/paste code, logic was repeated without any difference. Also and more importantly hydra service was relying on both hydra client and kratos client, which is against the single responsibility principle.

Having the controller layer rely on only one service is not a "practice", although in our codebase it appears like that. That would bring a bloated single service that has to do everything the controller (handler) needs.

It is perfectly normal/common to have a handler/controller orchestrate responses from multiple services, with each service doing "their own" singly responsible thing.

What you don't want is same-layer services being mutually dependant on one another.

pkg/kratos/handlers.go
```diff
@@ -137,42 +157,109 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+		a.mfaSettingsRedirect(w)
+		return
+	}
+
 	w.WriteHeader(http.StatusOK)
-	w.Write(resp)
+	_ = json.NewEncoder(w).Encode(flow)
+}
+
+func (a *API) shouldEnforceMFA(ctx context.Context, cookies []*http.Cookie) (bool, error) {
```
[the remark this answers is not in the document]
BarcoMasile: I thought about this, I prefered to place it in the controller layer because this afffects the request processing directly.
It wouldn't change much either way

BarcoMasile:
> Not many comments from a code perspective, but I have a question: Why are we checking the kratos session on the consent endpoint? I would think that kratos should be involved only in the login flow.

So I'm not able 100% to reply on this since that logic was already there and with this being the first touch on login-ui for me, I am not 100% confident on the whys etc, but from what I can get basically say that before my changes we needed for the identity ID for the `AcceptConsent` invocation.
After my changes the session is also needed to distinguish between external provider login and local idp login, and for the identity ID in the `HasTOTPAvailable` invocation

BarcoMasile:
In a chat with @nsklikas and @natalian98 , he pointed out a better solution to avoid mixing MFA enforcement and `GetConsent` logic. The improvement consist in checking MFA enforment needs **before** actually accepting the consent through hydra, since previously the constent would be accepted as part of the `CreateLoginFlow` logic, which made the guard for MFA necessary also in the `GetConsent` handler.

Next push will provide that improvement.
Thanks @nsklikas

pkg/kratos/handlers.go
```diff
@@ -137,42 +165,118 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+	if !a.mfaEnabled {
+		return false, nil
+	}
+
+	session, _, err := a.service.CheckSession(ctx, cookies)
+	if err != nil {
+		if a.is40xError(err) {
+			a.logger.Debugf("check session failed, err: %v", err)
```
natalian98, line 221: if I understand correctly, this is not a real error because we let kratos redirect to the 2fa screen as specified in config instead of enforcing it?

README.md
```diff
@@ -53,13 +53,15 @@ At the moment the application is sourcing the following from the environment:
...
-- `KRATOS_PUBLIC_URL` - address of Kratos APIs
+- `KRATOS_PUBLIC_URL` - address of Kratos Public APIs
+- `KRATOS_ADMIN_URL` - address of Kratos Admin APIs
 - `HYDRA_ADMIN_URL` - address of Hydra admin APIs
 - `OPENFGA_API_SCHEME` - the OpenFGA API scheme
 - `OPENFGA_API_HOST` - the OpenFGA API host name
 - `OPENFGA_STORE_ID` - the OpenFGA store ID to use
 - `OPENFGA_MODEL_ID` - the OpenFGA model ID to use. If not specified, a new
   model will be created
+- `MFA_ENABLED` - wether MFA is enabled and enforced, defaults to true
```
natalian98, line 64:
```suggestion
- `MFA_ENABLED` - whether MFA is enabled and enforced, defaults to true
```

natalian98 (approved):
works well :tada:

nsklikas (dismissed):

pkg/kratos/handlers.go
BarcoMasile: Yes, because we can't access the `whoami` endpoint so we get a 401 or 403, in that case we don't have to enforce mfa, because we have to do it after the actual login happen and we get the session to check the necessary stuff

README.md
BarcoMasile: Fixed

natalian98 (approved):

nsklikas (approved):

PR #269, opened 2024-08-28
Support MFA with backup codes

natalian98:
This PR adds support for lookup secrets (backup codes) as a fallback method for TOTP multi-factor auth.
--- ---
Setting up the codes (`/ui/setup_backup_codes`):
image
image
Print/download:
image
image

image

Sign in with a backup code:
image
image
Error messages:
image
If user logged in using a backup code and has 3 or less of them left, the following screen will be displayed, offering to generate a new set:
image


pkg/kratos/handlers.go
```diff
@@ -206,10 +207,56 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		if a.is40xError(err) {
+			a.logger.Debugf("check session failed: %v", err)
+			return false, nil
+		}
+
+		return false, err
+	}
+
+	aal2AuthenticationMethod := session.AuthenticationMethods[1].Method
```
BarcoMasile, line 245: question: is index 1 here because we're at the second step of the mfa-enforced login?

pkg/kratos/service.go
```diff
@@ -657,6 +696,47 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
...
+	identity, _, err := s.kratosAdmin.IdentityApi().
+		GetIdentity(ctx, id).
+		IncludeCredential([]string{"lookup_secret"}).
+		Execute()
+
+	if err != nil {
+		return false, err
+	}
+
+	lookupCredentials := identity.GetCredentials()["lookup_secret"].Config["recovery_codes"]
```
BarcoMasile, line 710: possibile issue: do we know for certain that `"recovery_codes"` key will always be available?
If that's not the case, I'll put my suggestion here:

```go
lookUpSecret, ok := identity.GetCredentials()["lookup_secret"]
if !ok {
    // handle case key is not present
}

lookupCredentials, ok := lookUpSecret.Config["recovery_codes"]
if !ok {
    // handle case key is not present
}
```

pkg/kratos/service.go
```diff
@@ -657,6 +696,47 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
...
+	if err != nil {
+		s.logger.Errorf("Marshalling to json failed: %s", err)
+		return false, err
+	}
+
+	lookupSecrets := new(LookupSecrets)
+	if err := json.Unmarshal(jsonbody, &lookupSecrets); err != nil {
+		s.logger.Errorf("Unmarshalling failed: %s", err)
+		return false, err
+	}
```
BarcoMasile, line 722: suggestion(non-blocking): how about switching to the `json.NewDecoder` approach, so it's leaner ?

pkg/kratos/service.go
```diff
@@ -657,6 +696,47 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
...
+	unusedCodes := 0
+	for _, code := range *lookupSecrets {
+		if code.UsedAt.IsZero() {
+			unusedCodes += 1
+		}
+	}
+
+	if unusedCodes > 3 {
+		return false, nil
+	}
```
BarcoMasile, line 733: suggestion: this is great already, but I suggest you create at least a function to isolate this specific logic of determining if codes available are enough or not. I'm referring to lines 731-733

ui/components/Flow.tsx
```diff
@@ -21,7 +21,14 @@ export type Values = Partial<
...
 
-export type Methods = "oidc" | "password" | "code" | "totp" | "webauthn";
+export type Methods =
+  | "oidc"
+  | "password"
+  | "code"
+  | "totp"
+  | "webauthn"
+  | "lookup"
+  | "lookup_secret";
```
BarcoMasile, line 31: question: is "lookup" an alias for "lookup_secret"?

ui/components/NodeInputSubmit.tsx
```diff
@@ -33,39 +33,55 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
+        ) : (
+          label
+        )}
+      </Button>
+      {showBackupLink && (
+        <Button
+          appearance="link"
+          type="button"
+          onClick={() =>
+            (window.location.href = window.location.href + "&use_backup_code=1")
```
BarcoMasile, line 81: question: I would expect a `true` value instead of `1`, just curious to know why we have an int here

ui/components/NodeText.tsx
```diff
@@ -18,23 +18,50 @@ const Content: FC<Props> = ({ attributes }) => {
...
+                  const element = document.createElement("a");
+                  const file = new Blob([text], {
+                    type: "text/plain",
+                  });
+                  element.href = URL.createObjectURL(file);
+                  element.download = "backup-codes.txt";
+                  document.body.appendChild(element);
+                  element.click();
+                  document.body.removeChild(element);
+                }}
```
BarcoMasile: suggestion: I have to ask why inlining this whole function here? Why not relying on `useCallback` ?
Same applies at line 53

ui/components/NodeText.tsx
```diff
@@ -29,21 +31,27 @@ const Content: FC<Props> = ({ attributes }) => {
...
-                  const element = document.createElement("a");
-                  const file = new Blob([text], {
-                    type: "text/plain",
-                  });
-                  element.href = URL.createObjectURL(file);
-                  element.download = "backup-codes.txt";
-                  document.body.appendChild(element);
-                  element.click();
-                  document.body.removeChild(element);
+                onClick={async () => {
```
BarcoMasile, line 44: same suggestoin here to use `useCallback`

BarcoMasile (commented):
Great work overall, some suggestions

ui/components/Flow.tsx
natalian98: Only `lookup_secret` is valid, I removed the `lookup`.

pkg/kratos/handlers.go
```diff
@@ -279,6 +326,27 @@ func (a *API) mfaSettingsRedirect(w http.ResponseWriter) {
 	)
 }
 
+func (a *API) lookupSecretsSettingsRedirect(w http.ResponseWriter, flowId string) {
+	redirectUrl, err := url.Parse(path.Join("/", a.contextPath, "/ui/backup_codes_regenerate?flow="+flowId))
```
nsklikas, line 330: nit-pick: use the `UI` constant found in the `ui` package instead of hardcoding the `ui` path?

pkg/kratos/handlers.go
```diff
@@ -279,6 +326,27 @@ func (a *API) mfaSettingsRedirect(w http.ResponseWriter) {
...
+	if err != nil {
+		err = fmt.Errorf("unable to build backup codes redirect path, possible misconfiguration, err: %v", err)
+		a.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	redirect := redirectUrl.String()
+	errorId := "regenerate_backup_codes"
```
nsklikas, line 339: suggestion: Move the `"regenerate_backup_codes"` to a global constant.

pkg/kratos/handlers.go
```diff
@@ -206,10 +207,56 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+
+		a.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if shouldRegenerateBackupCodes {
+		a.lookupSecretsSettingsRedirect(w, flowId)
+		return
+	}
```
nsklikas, line 222: question: why do we set the cookies before we apply this logic? If we get an error, we will return a 500 but still initialize a session, which feels a little counter intuitive for me.

pkg/kratos/handlers.go
```diff
@@ -206,10 +207,56 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
 	}
 
+	shouldRegenerateBackupCodes, err := a.shouldRegenerateBackupCodes(r.Context(), cookies)
+	if err != nil {
+		err = fmt.Errorf("error when checking backup codes: %v", err)
+
+		a.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
```
nsklikas, line 217: Not sure if I understand correctly, so please correct me if I'm wrong:

This endpoint may be called multiple times during a user's login flow, once for the username/password, once for OTP, once for webauthn. This means that this logic will run multiple times and if I don't want to print new codes (which is not unlikely) I will have to decline this every time. 

I have not run it to see if this is the actual behavior, so I may be missing something.

nsklikas (commented):
code looks fine (didn't look to much into the frontend), but I have a some questions and some nit-picks

pkg/kratos/handlers.go
natalian98: The user will only be offered to regenerate the codes if the method used to perform MFA was lookup secret and there are 3 or less codes left.
The screen is not presented multiple times (there are checks in the `shouldRegenerateBackupCodes` handler to prevent this) so the user can only decline it once.

pkg/kratos/service.go
natalian98: Agreed in private that it's not worth to change it to maintain readability

pkg/kratos/service.go
natalian98: Technically yes, because this check is only executed if the user logged in with `lookup_secret` method, meaning the identity has recovery codes. But I added the additional checks in case anything goes wrong.

pkg/kratos/handlers.go
nsklikas: Ah, I missed that thanks. What if we have more than 2fa though? Is this possible to happen (passkey as 3rd fa or something like that)

natalian98: Passkey can only be used as the first factor (this is restricted by kratos settings). No matter how many factors are there, this screen will only be shown once when a backup code is used.

ui/components/NodeInputSubmit.tsx
edlerd: changing it to true.

ui/components/NodeText.tsx
edlerd: Extracted and wrapped with use callback.

BarcoMasile (approved):
I'll approve so we can merge, there are still a few inline labmdas to swtich to the `useCallback` approach, but that'll be at a different time
Thank you

nsklikas (dismissed):

pkg/kratos/service.go
```diff
@@ -657,6 +697,57 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
 	return ok, nil
 }
 
+func (s *Service) HasNotEnoughLookupSecretsLeft(ctx context.Context, id string) (bool, error) {
```
shipperizer, line 700: this method has no tests

pkg/kratos/handlers.go
```diff
@@ -199,17 +203,69 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+
 	setCookies(w, cookies)
 
 	if shouldEnforceMfa {
```
shipperizer, line 217: not part of this pr, but there are not tests covering this if clause

mind adding an issue in gh?

natalian98: Sure
https://github.com/canonical/identity-platform-login-ui/issues/271

pkg/kratos/handlers.go
```diff
@@ -279,6 +335,27 @@ func (a *API) mfaSettingsRedirect(w http.ResponseWriter) {
 	)
 }
 
+func (a *API) lookupSecretsSettingsRedirect(w http.ResponseWriter, flowId string) {
+	redirectUrl, err := url.Parse(path.Join("/", a.contextPath, ui.UI, "/backup_codes_regenerate?flow="+flowId))
```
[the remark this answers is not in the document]
natalian98: Yes, because it's a subpath

pkg/kratos/service.go
natalian98: Tests added

BarcoMasile (approved):

pkg/kratos/handlers.go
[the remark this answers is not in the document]
shipperizer: subpath as front-end double slash path?

shipperizer (approved):

Issue #276, opened 2024-09-06
Unknown error code on invalid totp


nsklikas:
If I try to input a otp that does not have the right length (6), then I get this error:
![image](https://github.com/user-attachments/assets/cf7e13af-e09a-458c-90c8-c6ce678245d1)

natalian98:
https://www.ory.sh/docs/kratos/concepts/ui-user-interface#length-must-be--5-but-got-3-4000003

nsklikas:
Since we have an exhaustive list of errors, shouldn't we create a mapping for the ones that make sense and just return `Server Error` for the rest?

PR #280, opened 2024-09-11
Handle more kratos error codes

natalian98:
fixes https://github.com/canonical/identity-platform-login-ui/issues/276

### DIFF
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -20,17 +20,21 @@ import (
 )
 
 const (
-	IncorrectCredentials     = 4000006
-	InactiveAccount          = 4000010
-	InvalidRecoveryCode      = 4060006
-	RecoveryCodeSent         = 1060003
-	InvalidProperty          = 4000002
-	InvalidAuthCode          = 4000008
-	MissingSecurityKeySetup  = 4000015
-	BackupCodeAlreadyUsed    = 4000012
-	InvalidBackupCode        = 4000016
-	MissingBackupCodesSetup  = 4000014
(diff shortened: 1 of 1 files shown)


shipperizer (approved):
would be good to eventually get a test for the ui errors

shipperizer (approved):

pkg/kratos/service.go
```diff
@@ -417,8 +426,11 @@ func (s *Service) getUiError(responseBody io.ReadCloser) (err error) {
 		err = fmt.Errorf("invalid backup code")
 	case MissingBackupCodesSetup:
 		err = fmt.Errorf("login with backup codes unavailable")
+	case PasswordIdentifierSimilarity:
+		err = fmt.Errorf("password can not be similar to the email")
 	default:
-		err = fmt.Errorf("unknown kratos error code: %v", errorCode)
+		s.logger.Debugf("Unknown kratos error code: %v", errorCode)
```
nsklikas, line 432: personal preference, feel free to disregard: I don't think `debug` is the correct log level for this, IMO I would want to see the error code even in a production env

natalian98:
> I am approving this to move forward. BUT as I mentioned in the issue as well, I think we will need to go through all the Kratos errors and create a (nearly) exhaustive mapping. Adding an error code every time one pops up does not scale very well.

I went through the list of codes. Some of them are already mitigated by validation done on frontend (e.g. password requirements), others make no value to the end user.

shipperizer (approved):

nsklikas (approved):

PR #284, opened 2024-09-20
Delete session before recovery

nsklikas:
IAM-1044

This is hacky, but I don't think there is a better way to implement it.

Fixes https://github.com/canonical/kratos-operator/issues/259

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"path"
 	"strconv"
+	"time"
 
 	"github.com/go-chi/chi/v5"
 	client "github.com/ory/kratos-client-go"
@@ -17,6 +18,7 @@ import (
 )
 
 const RegenerateBackupCodesError = "regenerate_backup_codes"
+const KRATOS_SESSION_COOKIE_NAME = "ory_kratos_session"
 
 type API struct {
 	mfaEnabled  bool
@@ -456,7 +458,12 @@ func (a *API) handleCreateRecoveryFlow(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	setCookies(w, cookies)
+	setCookies(w, cookies, KRATOS_SESSION_COOKIE_NAME)
+	// We delete any active Kratos sessions. If there were any active Kratos sessions,
+	// recovery wouldn't be needed.
+	// See https://github.com/canonical/kratos-operator/issues/259 for more info.
+	a.deleteKratosSession(w)
+
 	w.WriteHeader(http.StatusOK)
 	w.Write(resp)
 }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -53,6 +53,7 @@ At the moment the application is sourcing the following from the environment:
   `log.txt`. **Make sure application user has permissions to write**.
 - `PORT` - HTTP server port, defaults to `8080`
 - `BASE_URL` - the base url that the application will be running on
+- `KRATOS_SESSION_COOKIE_NAME` - the kratos session cookie name, defaults to `ory_kratos_session`
```
shipperizer, line 56: i'd say let's make it a `const` and job on, it wont be changeable or at least, we won't be changing this

pkg/kratos/handlers.go
```diff
@@ -560,12 +567,29 @@ func (a *API) handleCreateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func NewAPI(service ServiceInterface, mfaEnabled bool, baseURL string, logger logging.LoggerInterface) *API {
+func (a *API) deleteKratosSession(w http.ResponseWriter) {
+	// To delete the session we delete the kratos session cookie.
+	// This is hacky as it does not call the Kratos API and is likely to break on
+	// a new Kratos version, but there is no easy way to delete the session
+	// from the Kratos API
+	c := &http.Cookie{
+		Name:     a.kratosSessionCookieName,
```
shipperizer, line 576:
```
const KRATOS_COOKIE_NAME = "whatever"
```

picking up on on the point above

```suggestion
		Name:     KRATOS_COOKIE_NAME,
```

pik4ez-canonical (dismissed):

pik4ez-canonical:
For me looks like acceptable level of hackiness at this stage.

natalian98:
While this solves the issue from https://github.com/canonical/kratos-operator/issues/259, it breaks another case:
If you log in using password, then without inserting the totp code you go to `/ui/reset_password`, that error appears:
image

pkg/kratos/handlers.go
```diff
@@ -580,6 +603,18 @@ func NewAPI(service ServiceInterface, mfaEnabled bool, baseURL string, logger lo
 	return a
 }
 
+func setCookiesWithExclude(w http.ResponseWriter, cookies []*http.Cookie, exclude []string) {
```
shipperizer, line 606:
```suggestion
func setCookies(w http.ResponseWriter, cookies []*http.Cookie, exclude ...string) {
```

I'd say to override setCookies with a variadic so we keep it the same and should not break backwards compatibility

nsklikas: ahh thanks. I knew there had to be a better way to do this, but I suck at go....

shipperizer (approved):

PR #285, opened 2024-09-23
Display error when the login flow is missing or invalid.

### LINKED ISSUE #281: Page does not load, if it fails to get login flow
If you try to access the login page with a wrong `flow` parameter, you will see a white page, e.g. try to go to https://iam.dev.canonical.com/stg-identity-jaas-dev-login-ui/ui/login?flow=57660e6b-fd54-491a-9dda-02d76641.

This can happen if flow expires as well. IMHO the page should load and show an error to the user.

edlerd:
# Done

- Display error when the login flow is missing or invalid.

fixes #281

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -37,6 +37,11 @@ const Login: NextPage = () => {
     use_backup_code: useBackupCode,
   } = router.query;
 
+  const redirectToErrorPage = () => {
+    const idParam = flowId ? `?id=${flowId.toString()}` : "";
+    window.location.href = `./error${idParam}`;
+  };
+
   useEffect(() => {
     // If the router is not ready yet, or we already have a flow, do nothing.
     if (!router.isReady || flow) {
@@ -48,7 +53,8 @@ const Login: NextPage = () => {
       kratos
         .getLoginFlow({ id: String(flowId) })
         .then((res) => setFlow(res.data))
-        .catch(handleFlowError("login", setFlow));
+        .catch(handleFlowError("login", setFlow))
+        .catch(redirectToErrorPage);
       return;
     }
 
@@ -67,7 +73,8 @@ const Login: NextPage = () => {
         }
         setFlow(data);
       })
-      .catch(handleFlowError("login", setFlow));
+      .catch(handleFlowError("login", setFlow))
+      .catch(redirectToErrorPage);
   }, [
     flowId,
     router,


ui/pages/login.tsx
```diff
@@ -48,7 +48,10 @@ const Login: NextPage = () => {
       kratos
         .getLoginFlow({ id: String(flowId) })
         .then((res) => setFlow(res.data))
-        .catch(handleFlowError("login", setFlow));
+        .catch(handleFlowError("login", setFlow))
+        .catch(() => {
+          window.location.href = "./oidc_error";
```
nsklikas, line 53: question: I don't know about the latest changes in the login UI, but I thought that this page was used only to display errors coming from Hydra (related to the OIDC flow). I would think that we would redirect the user to `error` instead, am I missing something?

nsklikas (commented):
I am not sure what the behavior of the login UI should be on this one, redirecting the user to the error page is one approach but it is not great from a UX perspective. If I leave my browser on the login page and refresh it after one day, I will get an error. 
Instead I think that creating a new flow and using that instead on the URL would be a better approach, as we would just load the login page again (but tbh the user will be unable to do anything after logging in, as they probably won't be able to go back to Hydra and we don't have a menu for the login UI)

What do you think? I am not great with UX, so maybe we should get the input of someone better at this.

edlerd:
> I am not sure what the behavior of the login UI should be on this one, redirecting the user to the error page is one approach but it is not great from a UX perspective. If I leave my browser on the login page and refresh it after one day, I will get an error. Instead I think that creating a new flow and using that instead on the URL would be a better approach, as we would just load the login page again (but tbh the user will be unable to do anything after logging in, as they probably won't be able to go back to Hydra and we don't have a menu for the login UI)
> 
> What do you think? I am not great with UX, so maybe we should get the input of someone better at this.

There are different root problems, we are trying to solve. I would like to differentiate:

1. We don't have a flow, or the flow-id we have is invalid. In this case, all we can do is show an error message. I agree the message should be enhanced to ask the user to start a new login process.

2. The flow expired. In this case, we should start a new one if possible. Question: Is this feasible and do we have enough information to start a new flow from the login UI? If the answer is no, we can also only show an error message.

ui/pages/login.tsx
edlerd: Agreed, this should be a dedicated error message and not using the oidc error page. Will change this shortly.

nsklikas:
> There are different root problems, we are trying to solve. I would like to differentiate:
> 
>     1. We don't have a flow, or the flow-id we have is invalid. In this case, all we can do is show an error message. I agree the message should be enhanced to ask the user to start a new login process.
> 
>     2. The flow expired. In this case, we should start a new one if possible. Question: Is this feasible and do we have enough information to start a new flow from the login UI? If the answer is no, we can also only show an error message.

In the case we don't have a flow, a new one is created (see https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/login.tsx#L56). 

I think the differentiation should be:

1. No flow, this is the case every time someone is redirected to the login UI from Hydra
2. Expired or invalid flow, I am not sure if we can differentiate between the 2 and I am not sure we would gain much even if could

IMHO the main problem is that we don't have a landing page to redirect the user after they login, which means that it doesn't make much sense to create a new flow the one provided is expired or invalid. I guess in that case we can just show an generic error, like `The login session is expired or invalid, try again`, and when we get a proper landing page we can start creating a new flow and redirecting the user there.

ui/pages/login.tsx
```diff
@@ -67,7 +73,8 @@ const Login: NextPage = () => {
         }
         setFlow(data);
       })
-      .catch(handleFlowError("login", setFlow));
+      .catch(handleFlowError("login", setFlow))
+      .catch(redirectToErrorPage);
```
nsklikas, line 77: If the flow creation fails, then I would expect that there would be no flow_id to set. What will happen if we redirect to the error page with an empty flow_id?

edlerd: We will fall back to a generic message: `An error occurred. Try signing in again.` in case there is no id set or if we can't load the id specified.

edlerd:
> IMHO the main problem is that we don't have a landing page to redirect the user after they login, which means that it doesn't make much sense to create a new flow the one provided is expired or invalid. I guess in that case we can just show an generic error, like `The login session is expired or invalid, try again`, and when we get a proper landing page we can start creating a new flow and redirecting the user there.

Yes, exactly. We might not have enough information to start a new flow with the correct URL to redirect to after login is successful. Is there a way to start a new flow with the correct redirect URL from an expired flow? That was the 2nd case I mentioned above, that we might be able to do better than just show an error message. Showing the error page obviously works in all cases. But we might be able to do better in that specific case of a flow being expired -- if we can start a new valid flow that "just works" instead of the expired one.

nsklikas:
> Is there a way to start a new flow with the correct redirect URL from an expired flow?

AFAIK no

Right now if you go to `/ui/login`, you will see an error page
image


UNLESS you have an active session and you have not set up the authenticator, in that case you will be asked to register the authenticator and then be redirected to the `setup_complete` page. Not sure how this will behave after https://github.com/canonical/identity-platform-login-ui/pull/287 is merged.

nsklikas (approved):

BarcoMasile (approved):

PR #286, opened 2024-09-23
Hide passkey option in case it is disabled

### LINKED ISSUE #277: Uncorrect method sent in the payload when `webauthn` is disabled
If `webauthn` config is disabled and you attempt to sign in using a security key, UI displays a message "invalid password". This is because it sends an invalid payload (with `password` method), so kratos renders [this](https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/pkg/kratos/service.go#L408) error message as it doesn't recognize it.

image

If `webauthn` is enabled and you attempt to sign in with a user that doesn't have any webauthn credentials, it will display the correct message because the method in the payload is correct:
image

edlerd:
# Done

- Hide passkey option in webauthn is not configured

fixes #277 WD-15073

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -163,9 +163,12 @@ const Login: NextPage = () => {
     };
   };
 
+  let isWebauthn = false;
+  const supportsWebauthn = flow?.ui.nodes.some(
+    (node) => node.group === "webauthn",
+  );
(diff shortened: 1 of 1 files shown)


nsklikas (approved):
I removed the webauth config from the `kratos.yml` and the `Login with passkey` button disappeared (as expected). LGTM, thanks for the fix

shipperizer (approved):

PR #287, opened 2024-09-24
Fix redirect on first login

### LINKED ISSUE #275: No redirection on first login
By deploying the bundle found on https://github.com/canonical/iam-bundle/pull/273, if I create an admin account through the kratos juju cli (`juju run kratos/0 create-admin-account email=a@b.com password=123a username=aba`) and I try to login via Hydra, I will be asked to register my authenticator and then I will end up in the `setup_complete` page instead of being redirected back to Hydra.

nsklikas:
IAM-1039

The issue is fixed by passing the original return to, to the settings flow. I moved some stuff around, to more clearly separate the responsibilities of the frontend and the backend. 

There is some reload happening in the `setup_secure` page when you get there (you momentarily see the spinning thingy). Not sure how to fix it (@edlerd please have a look)

SInce Natalia is not here this week @edlerd and @BarcoMasile please try it out, as I may have broken something

### DIFF
--- a/ui/pages/setup_secure.tsx
+++ b/ui/pages/setup_secure.tsx
@@ -44,26 +44,27 @@ const SetupSecure: NextPage = () => {
     // Otherwise we initialize it
     kratos
       .createBrowserSettingsFlow({
-        returnTo: returnTo ? String(returnTo) : undefined,
+        returnTo: returnTo
+          ? returnTo.toString()
+          : window.location.pathname.replace("setup_secure", "setup_complete"),
       })
-      .then(({ data }) => {
-        if (data.request_url !== undefined) {
-          const pwParam = pwChanged
-            ? `&pw_changed=${pwChanged.toString()}`
-            : "";
-          window.location.href = `./setup_secure?flow=${data.id}${pwParam}`;
-          return;
-        }
+      .then(async ({ data }) => {
+        const pwParam = pwChanged ? pwChanged.toString() : "false";
+        router.query.flow = data.id;
+        router.query.pw_changed = pwParam;
+
+        await router.replace(
+          {
+            pathname: window.location.pathname,
+            query: router.query,
+          },
+          undefined,
+          { shallow: true },
+        );
         setFlow(data);
       })
       .catch(handleFlowError("settings", setFlow))
       .catch(async (err: AxiosError<string>) => {
-        if (err.response?.data.trim() === "Failed to create settings flow") {
-          setFlow(undefined);
-          window.location.href = "./login";
-          return;
-        }
-
         return Promise.reject(err);
       });
   }, [flowId, router, router.isReady, returnTo, flow]);
@@ -83,15 +84,12 @@ const SetupSecure: NextPage = () => {
           },
         })
         .then(({ data }) => {
-          if (flow?.state === "success") {
-            window.location.href = "./setup_complete";
-          }
           if ("redirect_to" in data) {
             window.location.href = data.redirect_to as string;
             return;
           }
-          if (flow?.return_to) {
-            window.location.href = flow.return_to;
+          if (data?.return_to) {
+            window.location.href = data?.return_to;
             return;
           }
           window.location.href = "./error";
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -75,7 +75,7 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 		shouldEnforceMfa, err = a.shouldEnforceMFAWithSession(r.Context(), session)
 
 		if shouldEnforceMfa {
-			a.mfaSettingsRedirect(w)
+			a.mfaSettingsRedirect(w, returnTo)
 			return
 		}
 
@@ -217,7 +217,7 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 	setCookies(w, cookies)
 
 	if shouldEnforceMfa {
-		a.mfaSettingsRedirect(w)
+		a.mfaSettingsRedirect(w, *loginFlow.ReturnTo)
 		return
 	}
 
@@ -317,8 +317,9 @@ func (a *API) is40xError(err error) bool {
 	return false
 }
 
-func (a *API) mfaSettingsRedirect(w http.ResponseWriter) {
+func (a *API) mfaSettingsRedirect(w http.ResponseWriter, returnTo string) {
 	redirect, err := url.JoinPath("/", a.contextPath, "/ui/setup_secure")
+
 	if err != nil {
 		err = fmt.Errorf("unable to build mfa redirect path, possible misconfiguration, err: %v", err)
 		a.logger.Error(err.Error())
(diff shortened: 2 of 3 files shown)


ui/pages/setup_secure.tsx
```diff
@@ -41,32 +41,36 @@ const SetupSecure: NextPage = () => {
       return;
     }
 
+    const returnTo = _returnTo
+      ? _returnTo.toString()
+      : window.location.pathname + `/../setup_complete`;
```
nsklikas, line 46: This is hacky, @edlerd can we refactor this?

edlerd: Why not define a `const setupComplete = window.location.pathname + `/../setup_complete`;` and keep the previous line 47 below a bit more as it was: `returnTo: returnTo ? String(returnTo) : setupComplete`?

nsklikas: I removed the extra variable, thanks for the catch.

But my question still remains about how we should construct the path to the `setup_complete` page.

edlerd: Doesn't the relative path as `./setup_complete` should work just fine?

nsklikas: No, the current path will be something like `http://login-ui.com/ui/setup_secure`, this means that the relative path of `./setup_complete` will result to `http://login-ui.com/ui/setup_secure/setup_complete`

The current code just happened to work, because certain parts of the code were unreachable. 
For example:
- https://github.com/canonical/identity-platform-login-ui/pull/287/files#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9L87 would never be executed because the condition:
```
if (flow?.state === "success")
```
would never be true because `setFlow` was never called to update the flow with the result of `updateSettingsFlow`.
- The same is true for https://github.com/canonical/identity-platform-login-ui/pull/287/files#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9L93-L94, the condition:
```
if (flow?.return_to)
```
was never true
- Also I think that [this](https://github.com/canonical/identity-platform-login-ui/pull/287/files#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9L50) was always true, as the flow always has a `request_url`, which means that the `setFlow` at https://github.com/canonical/identity-platform-login-ui/pull/287/files#diff-2e493fbf6fdb5973f71ece440a2849e4b4765ab23f84cfdd4a9186c8746f0cc9L57 was unreachable.

edlerd: > No, the current path will be something like `http://login-ui.com/ui/setup_secure`, this means that the relative path of `./setup_complete` will result to `http://login-ui.com/ui/setup_secure/setup_complete`

That is surprising. Is the current url maybe `http://login-ui.com/ui/setup_secure/` with a trailing slash? Then we should remove that trailing slash and the redirect with the relative path should work fine.

Another option is we could traverse the path upwards until we find `/ui/` and treat that as the root. But this only works under the assumption the ui will always be hosted under a `/ui` prefix. I'd prefer the 1st option as it seems easier to me.

nsklikas: But this is not a redirect, this is concatenation:
```
"http://login-ui.com/ui/setup_secure/./setup_complete" = "http://login-ui.com/ui/setup_secure/setup_complete" 
```

And:
```
"http://login-ui.com/ui/setup_secure./setup_complete" = Invalid
```

We construct a URL and give it to Kratos and tell to Kratos that the user should be redirected there AFTER they finish their flow. You don't know at which URL the user will be when they finish their flow (so a relative path does not make much sense).

In your example we would have to tell Kratos to redirect the user to a relative path, but I am not sure if this works. I am pretty sure Kratos won't allow you to set a relative path (e.g. `./setup_complete`) as the return_to URL.

edlerd: I understand what you meant above now. This is not a redirect, but a URL to be used later as we initiate the kratos flow.

Is the path guaranteed to include the `/ui/` prefix? If yes, we could parse the window.location.pathname and remove anything after the last occurrence of `/ui/` to get a base URL. Then we can append `setup_complete` to the base URL.

lukasSerelis:
Pasting reply on this issue too for UX requirements 
> It'd be best from UX point of view to add a confirmation screen before the automatic redirect to the application. So the flow should be something like this: image Or in screens it would be
> 
> 1. Trying to sign in for the first time
>    image
> 2. Required TOTP setup
>    image
> 3. Setup complete, message to confirm that action has been successful (also good place to show any errors if some occur). And automatic redirect within 3 seconds.
>    image
> 4. Landing on the 3rd party application page
>    image

ui/pages/setup_secure.tsx
edlerd: Actually this is way simpler, we are on the `setup_secure` page module here, so just replace `setup_secure` with `setup_secure` in the current pathname should work well.

nsklikas: Thanks, I updated it. I don't love that I have to hard-code the file name in the code, but it works

ui/pages/setup_secure.tsx
```diff
@@ -21,7 +21,7 @@ const SetupSecure: NextPage = () => {
   // Get ?flow=... from the URL
   const router = useRouter();
   const {
-    return_to: returnTo,
+    return_to: _returnTo,
```
edlerd, line 24:
```suggestion
    return_to,
```
I think we can keep it simpler and avoid the aliasing here?

ui/pages/setup_secure.tsx
```diff
@@ -20,11 +20,7 @@ const SetupSecure: NextPage = () => {
 
   // Get ?flow=... from the URL
   const router = useRouter();
-  const {
-    return_to: returnTo,
-    flow: flowId,
-    pw_changed: pwChanged,
-  } = router.query;
+  const { return_to, flow: flowId, pw_changed: pwChanged } = router.query;
```
edlerd, line 23:
```suggestion
  const { return_to: returnTo, flow: flowId, pw_changed: pwChanged } = router.query;
```
Sorry, my previous comment was ambiguous, what I meant is to keep the old alias `returnTo` in favour of changing it to `_returnTo`. We like the camlCase and dislike snake_case in tsx code.

ui/pages/setup_secure.tsx
```diff
@@ -44,29 +40,30 @@ const SetupSecure: NextPage = () => {
     // Otherwise we initialize it
     kratos
       .createBrowserSettingsFlow({
-        returnTo: returnTo ? String(returnTo) : undefined,
+        returnTo: return_to
+          ? return_to.toString()
+          : window.location.pathname.replace(`setup_secure`, `setup_complete`),
```
edlerd, line 45:
```suggestion
          : window.location.pathname.replace('setup_secure', 'setup_complete'),
```
These are static strings, no need for a template

ui/pages/setup_secure.tsx
```diff
@@ -44,29 +40,30 @@ const SetupSecure: NextPage = () => {
...
+        router.query.pw_changed = pwParam;
+
+        await router.replace(
+          {
+            pathname: window.location.pathname,
+            query: router.query,
+          },
+          undefined,
+          { shallow: true },
+        );
```
edlerd, line 59: Did you test the flow with various combinations? Like 1. coming from the password change page or 2. hitting this page directly with an authenticated or 3. unauthenticated session? I think there was a reason we didn't use `router.replace` before, but I might be wrong.

edlerd (commented):
Mostly looks good, some non-blocking nitpicks and one concern below.

ui/pages/setup_secure.tsx
edlerd: Thanks for checking. In my opinion, (3) should redirect to log in. Which it does on staging, but then you get to the error page, because we don't allow login when arriving without a valid login challenge or flow ID. Not sure if we can fix it here, or should just report it as a bug?

edlerd: The (3) case was likely handled by the removed lines 61-66 below, maybe they should be restored?

nsklikas: I don't think that ever worked (because of what you said on your comment about the login challenge). I would like to remove it for now and we can fix it properly in another PR.

edlerd (approved):
LGTM, thanks for the fixes!

BarcoMasile (approved):

PR #292, opened 2024-09-27
fix show email in case it is a hidden field as disabled input with a change email link to restart the flow

### LINKED ISSUE #290: Sign in with security key not working if email is cached
**What is happening**
If you try to sign in but don't complete MFA and visit the initial sign in page again, it caches the email and only shows password and security key sign up options, even if security key is not set up. If the security key is not set up it's not putting an error message, it just keep redirecting back to the same page. 

**What should happen**
First of all it shouldn't obscure the email address that is cached, it's bad UX practice as the user might be trying to sign in to a different account, and you're making them guess which email has been cached. Display the email even thought it's been cached. Secondly, either display all methods of sign in or only supported ones, cause now you're doing neither.

**Preferred flow**
image

**Flow in screens**
1) User visits resource that is accessible via Login UI. UI displays all of possible sign in methods
image

2) User signs in with email and password. Login UI asks for MFA. User closes the tab
image

3) User visits resource that is accessible via Login UI. UI displays cached email with ability to edit email, and all sign in methods that this email is able to use (in this example, email has security key and google, but not Okta set up for this identity.)
image

**Compromise option**
Same flow, but display all of the possible sign in option, regardless of what options are possible. 
image

edlerd:
# Done
- fix show email in case it is a hidden field as disabled input with a change email link to restart the flow

fixes #290

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -30,6 +30,7 @@ type API struct {
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Delete("/api/kratos/self-service/clear-session", a.handleClearKratosSession)
 	mux.Post("/api/kratos/self-service/login", a.handleUpdateFlow)
 	mux.Get("/api/kratos/self-service/login/browser", a.handleCreateFlow)
 	mux.Get("/api/kratos/self-service/login/flows", a.handleGetLoginFlow)
@@ -40,6 +41,17 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Post("/api/kratos/self-service/settings", a.handleUpdateSettingsFlow)
 	mux.Get("/api/kratos/self-service/settings/browser", a.handleCreateSettingsFlow)
 	mux.Get("/api/kratos/self-service/settings/flows", a.handleGetSettingsFlow)
+}
+
+func (a *API) handleClearKratosSession(w http.ResponseWriter, _ *http.Request) {
+	http.SetCookie(w, &http.Cookie{
+		Name:   KRATOS_SESSION_COOKIE_NAME,
+		Value:  "",
+		Path:   "/",
+		MaxAge: -1,
+	})
+
+	w.WriteHeader(http.StatusOK)
 }
 
 // TODO: Validate response when server error handling is implemented
--- a/ui/components/NodeInputHidden.tsx
+++ b/ui/components/NodeInputHidden.tsx
@@ -1,7 +1,35 @@
 import { NodeInputProps } from "./helpers";
 import React, { FC } from "react";
+import { Input } from "@canonical/react-components";
 
-export const NodeInputHidden: FC<NodeInputProps> = ({ attributes }) => {
+export const NodeInputHidden: FC<NodeInputProps> = ({
+  node,
+  attributes,
+  setValue,
+}) => {
+  if (node.meta.label?.text === "Change e-mail") {
+    return (
+      <Input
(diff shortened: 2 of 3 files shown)


ui/pages/login.tsx
```diff
@@ -171,12 +184,38 @@ const Login: NextPage = () => {
...
+      renderFlow.ui.messages[0].id === 1010003;
+    if (isPasswordOnly) {
+      renderFlow.ui.nodes = renderFlow.ui.nodes.map((node) => {
+        if (
+          node.group !== "default" ||
+          (node.attributes as UiNodeInputAttributes).name !== "identifier"
+        ) {
+          return node;
+        }
+
```
BarcoMasile, line 205: can we wrap this 2 pieces of logic in two separate function with an explicative name?

BarcoMasile (approved):
LGTM, just a small request for code clarity

pkg/kratos/handlers.go
```diff
@@ -42,6 +43,17 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/kratos/self-service/settings/flows", a.handleGetSettingsFlow)
 }
 
+func (a *API) handleClearKratosSession(w http.ResponseWriter, _ *http.Request) {
+	http.SetCookie(w, &http.Cookie{
+		Name:   KRATOS_SESSION_COOKIE_NAME,
+		Value:  "",
+		Path:   "/",
+		MaxAge: -1,
+	})
```
nsklikas, line 52:
```suggestion
	a.deleteKratosSession(w)
```

ui/pages/login.tsx
```diff
@@ -253,6 +292,11 @@ const Login: NextPage = () => {
     <PageLayout title={title}>
       {flow ? <Flow onSubmit={handleSubmit} flow={renderFlow} /> : <Spinner />}
       {isWebauthn && <a href={flow?.return_to}>I want to use another method</a>}
+      {isPasswordOnly && (
+        <a href={`${window.location.href}&reset_email=1`}>
```
nsklikas, line 296: This causes the page to reload and adds this on the browser history, can't we avoid it?

ui/components/NodeInputHidden.tsx
```diff
@@ -1,7 +1,35 @@
...
+  if (node.meta.label?.text === "Change e-mail") {
+    return (
+      <Input
+        type="email"
+        labelClassName="password-label"
+        label={
+          <>
+            <span>E-Mail</span>
+            <a
+              href={`${window.location.href}&reset_email=1`}
```
nsklikas, line 19: question: This causes the page to reload and adds this on the browser history, can't we avoid it?

nsklikas (commented):
I don't really like the approach we take here:
- the button that says `Change e-mail`, seems weird/counter-intuitive to me as a user. 
- The logic for resetting aal1 feels very hacky and I think will be difficult for someone who does not have our context to extend.

We should have a look at what major providers do and copy them. For example on github (not the best example, I know) if you press back on the authorizer screen, you will be asked to enter your username/password again and if you do browser forward, you will again will be asked for totp.

I think for us the best flow would be to show the user a message that says something along the lines `Please validate that this is you: {email}` and underneath 2 buttons (one for accepting and one for declining). If they decline, then we delete the user session and we show the login screen as well. If they accept, then we set the aal param to `aal=aal2` in the query param. This will show the user the totp page again.

What do you think? cc @lukasSerelis @edlerd

edlerd:
Closing this PR in favour of #301 which includes the changes here.

PR #294, opened 2024-09-27
fix sign in without a challenge

### LINKED ISSUE #247: Sign in page not loading correctly
The login logic is currently tied to hydra. If you try to go to `/ui/reset_password` without a session, you'll get redirected to sign in but the page will be shown as loading:
image

This is due to missing `login_challenge` parameter value. Similar bug can be observed when going directly to http://localhost:4455/ui/login.

edlerd:
# Done
- fix sign in without a challenge
- i.e. on hitting the reset_password or other settings URLs directly

fixes #247

### DIFF
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -129,10 +129,12 @@ func (s *Service) CreateBrowserLoginFlow(
 		ReturnTo(returnTo).
 		Refresh(refresh).
 		Cookie(httpHelpers.
-		CookiesToString(cookies))
+			CookiesToString(cookies))
 
 	if loginChallenge != "" {
-		request.LoginChallenge(loginChallenge)
+		request = request.LoginChallenge(loginChallenge)
+	} else if loginChallenge == "" && returnTo == "" {
+		return nil, nil, fmt.Errorf("no return_to or login_challenge was provided")
 	}
 
 	flow, resp, err := request.Execute()
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -123,14 +123,19 @@ func (s *Service) CreateBrowserLoginFlow(
 	ctx, span := s.tracer.Start(ctx, "kratos.Service.CreateBrowserLoginFlow")
 	defer span.End()
 
-	flow, resp, err := s.kratos.FrontendApi().
+	request := s.kratos.FrontendApi().
 		CreateBrowserLoginFlow(ctx).
 		Aal(aal).
 		ReturnTo(returnTo).
-		LoginChallenge(loginChallenge).
 		Refresh(refresh).
-		Cookie(httpHelpers.CookiesToString(cookies)).
-		Execute()
+		Cookie(httpHelpers.
+		CookiesToString(cookies))
+
+	if loginChallenge != "" {
+		request.LoginChallenge(loginChallenge)
+	}
+
+	flow, resp, err := request.Execute()
 	if err != nil {
 		return nil, nil, err
 	}
--- a/pkg/kratos/service_test.go
+++ b/pkg/kratos/service_test.go
@@ -215,7 +215,7 @@ func TestAcceptLoginRequestFails(t *testing.T) {
 	}
 }
 
-func TestCreateBrowserLoginFlowSuccess(t *testing.T) {
+func TestCreateBrowserLoginFlowWithLoginChallengeSuccess(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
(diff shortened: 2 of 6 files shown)


pkg/kratos/service.go
```diff
@@ -123,14 +123,19 @@ func (s *Service) CreateBrowserLoginFlow(
...
-		Cookie(httpHelpers.CookiesToString(cookies)).
-		Execute()
+		Cookie(httpHelpers.
+		CookiesToString(cookies))
+
+	if loginChallenge != "" {
+		request.LoginChallenge(loginChallenge)
+	}
+
+	flow, resp, err := request.Execute()
```
nsklikas, line 138: We don't have a home page, this means that the user should not log in without a login challenge OR a return to. Let's add this logic to the code, I can make this change if you want

edlerd: Right, we should ensure either of the two are set. I was removing the need to pass the login challenge with this change. Because users might land on the "reset password" page, and should be able to login, which was previously failing. I think the reset password page would then be the return to address, so it should be fine as you suggested.

Please, can you add commits to my branch to enforce the requirement to have either the challenge or the return to present @nsklikas ?

ui/pages/setup_backup_codes.tsx
```diff
@@ -53,7 +53,7 @@ const SetupBackupCodes: NextPage = () => {
       .catch(async (err: AxiosError<string>) => {
         if (err.response?.data.trim() === "Failed to create settings flow") {
           setFlow(undefined);
-          window.location.href = "./login";
+          window.location.href = `./login?return_to=${window.location.pathname}`;
```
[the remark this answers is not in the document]
edlerd: I think in those cases here we want to actually redirect, so setting the location href would be preferred.

ui/pages/setup_passkey.tsx
```diff
@@ -53,7 +53,7 @@ const SetupPasskey: NextPage = () => {
       .catch(async (err: AxiosError<string>) => {
         if (err.response?.data.trim() === "Failed to create settings flow") {
           setFlow(undefined);
-          await router.push("./login");
+          await router.push(`./login?return_to=${window.location.pathname}`);
```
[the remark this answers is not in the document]
edlerd: right, will adjust this.

nsklikas (dismissed):

pkg/kratos/service.go
```diff
@@ -123,14 +125,21 @@ func (s *Service) CreateBrowserLoginFlow(
...
-		LoginChallenge(loginChallenge).
 		Refresh(refresh).
-		Cookie(httpHelpers.CookiesToString(cookies)).
-		Execute()
+		Cookie(httpHelpers.
+			CookiesToString(cookies))
+
+	if loginChallenge != "" {
+		request = request.LoginChallenge(loginChallenge)
+	} else if returnTo == "" {
```
shipperizer, line 138: is this `loginChallenge is "abc" and returnTo is not set`?

pkg/kratos/service.go
```diff
@@ -123,14 +125,21 @@ func (s *Service) CreateBrowserLoginFlow(
...
+		return nil, nil, ErrNoReturnToOrLoginChallenge
```
shipperizer, line 139:
```suggestion
		return nil, nil, fmt.Errorf("No return_to or login_challenge was provided")
```

i'd probably drop the const and tweak the tests

pkg/kratos/service_test.go
```diff
@@ -282,6 +282,109 @@ func TestCreateBrowserLoginFlowSuccess(t *testing.T) {
...
+
+	if f != nil {
+		t.Fatalf("expected flow to be %v not  %v", nil, f)
+	}
+	if c != nil {
+		t.Fatalf("expected cookies to be %v not  %v", nil, c)
+	}
+	if err != ErrNoReturnToOrLoginChallenge {
+		t.Fatalf("expected error to be %s, not %s", ErrNoReturnToOrLoginChallenge, err)
+	}
```
shipperizer, line 385: based on https://github.com/canonical/identity-platform-login-ui/pull/294/files#diff-5e4ba31969008b305d67765d5d4c9bd50d9bcc49571f1594e97d2b152ddf1378R139

```suggestion
	if err == nil {
		t.Fatalf("expected error to be not nil")
	}
```

pkg/kratos/service.go
nsklikas: why drop the const? I find it clearer to use only vars for generating tests, this allows you to gather them to a single place and imo makes it more manageable.

No problem to change it, just curious

pkg/kratos/service.go
nsklikas: No, it's when `loginChallenge == "" && returnTo == ""`

shipperizer: as discussed probably worth being more explicit with another separate `if`

pkg/kratos/service.go
shipperizer: main reason is we don't currently instantiate error messages this way in the codebase, i m happy to change it everywhere but i would keep it as a separate effort

 in terms of testing, fair point, at the same time, unless we are interested in the error message itself, what we should be checking is that the function errors (or at least that's my opinion)

shipperizer (approved):

nsklikas (approved):

edlerd:
@shipperizer or @nsklikas can you merge it?

Issue #279, opened 2024-09-10
`Failed to get login flow` when going back in the browser history


natalian98:
If you log in with email and password, login ui redirects to complete the 2fa. If instead of providing it you go back in the browser and attempt to sign in again, a `Failed to get login flow`  error will be displayed:
![image](https://github.com/user-attachments/assets/57488ba5-f701-4b5b-8fa6-9af9bcc1b613)
In login ui logs:
```
"message":"Error when getting login flow: 403 Forbidden\n"
```
The solution is to 1) clear the browser cookies or 2) go again to the original url which will redirect to insert the password again (the email is cached).

lukasSerelis:
Ideally this situation should not occur at all, and we could do this by providing better back-tracking ways in UI. Is there a way to give user a way back that wouldn't cause this error?

If we can not do that, we need to give a better error message in this case. Ideally not showing sign-in options at all and just showing an error screen as now it's indicating that something is wrong with inputs in the fields, whereas the real issue is that they need to clear cookies or start from the original sign in url.

So ideally the flow is either an accurate error or a button that doesn't break the UI 
![image](https://github.com/user-attachments/assets/f057d1db-531b-413b-8822-73d03a0a8eb4)

nsklikas:
I am unable to reproduce this issue for some time now, can you verify that it's still happening and provide more precise instructions?

lukasSerelis:
> I am unable to reproduce this issue for some time now, can you verify that it's still happening and provide more precise instructions?

https://iam.dev.canonical.com/stg-identity-jaas-dev-grafana-k8s/login~ happening here still

nsklikas:
This error happens on chromium, can't reproduce it on firefox

lukasSerelis:
Anything we can do for chromium?

edlerd:
This is due to [bfcache](https://web.dev/articles/bfcache). It can be disabled with a header `Cache-Control: "no-cache, no-store, must-revalidate"`, that we probably want to set on all responses of login ui.

lukasSerelis:
> This is due to [bfcache](https://web.dev/articles/bfcache). It can be disabled with a header `Cache-Control: "no-cache, no-store, must-revalidate"`, that we probably want to set on all responses of login ui.

Would this not break the flow of Enter email address -> Click reset password -> Email saved from starting screen -> Click reset password? @nsklikas @edlerd

edlerd:
[...]

PR #295, opened 2024-09-28
feat ui: link to use authenticator app instead in screen asking for a backup code

### LINKED ISSUE #293: MFA screens
Some issues with MFA screens. Currently if a user selects to verify identity with backup code it doesn't offer a way to change it to use Authenticator app instead. Also some error message text issues. 
Ideally the flow should be like this: 
image

Or in screens:
1) User signs in
image

2) Both Authenticator app and Backup code is set up
image

3) If a user click I want to use a backup code, we take them to this page, which shows the ability to go back to using authenticator app.
image

4) [Future cases] In a case where we support more methods for MFA than backup codes and TOTP, we should say "Use another method" and give users a selection like this
image
image

edlerd:
# Done

- feat ui: link to use authenticator app instead in screen asking for a backup code

fixes #293

### DIFF
--- a/ui/components/NodeInputSubmit.tsx
+++ b/ui/components/NodeInputSubmit.tsx
@@ -2,6 +2,7 @@ import { getNodeLabel } from "@ory/integrations/ui";
 import { Button } from "@canonical/react-components";
 import { NodeInputProps } from "./helpers";
 import React, { FC } from "react";
+import { useRouter } from "next/router";
 
 export const NodeInputSubmit: FC<NodeInputProps> = ({
   node,
@@ -10,6 +11,8 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
   disabled,
   dispatchSubmit,
 }) => {
+  const router = useRouter();
+
   const getProviderImage = (value: string) => {
     if (value.toLowerCase().startsWith("auth0")) {
       return "logos/Auth0.svg";
@@ -82,10 +85,17 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
           appearance="link"
           tabIndex={5}
           type="button"
-          onClick={() =>
-            (window.location.href =
-              window.location.href + "&use_backup_code=true")
-          }
+          onClick={() => {
+            const newQuery = { ...router.query, ["use_backup_code"]: "true" };
+            void router.push(
+              {
+                pathname: window.location.pathname,
+                query: newQuery,
+              },
+              undefined,
+              { shallow: true },
+            );
+          }}
         >
           Use backup code instead
         </Button>
@@ -95,12 +105,18 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
           appearance="link"
           tabIndex={5}
           type="button"
-          onClick={() =>
-            (window.location.href = window.location.href.replace(
-              "&use_backup_code=true",
-              "",
-            ))
-          }
+          onClick={() => {
+            const newQuery = { ...router.query };
+            delete newQuery.use_backup_code;
+            void router.push(
+              {
+                pathname: window.location.pathname,
+                query: newQuery,
+              },
+              undefined,
+              { shallow: true },
+            );
+          }}
         >
           Use authentication code instead
         </Button>
--- a/ui/components/NodeInputSubmit.tsx
+++ b/ui/components/NodeInputSubmit.tsx
@@ -36,6 +36,9 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
   const showBackupLink =
     (node.meta.label as unknown as { hasBackupLink: boolean })?.hasBackupLink ??
     false;
+  const showTotpLink =
+    (node.meta.label as unknown as { hasTotpLink: boolean })?.hasTotpLink ??
+    false;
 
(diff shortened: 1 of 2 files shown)


ui/components/NodeInputSubmit.tsx
```diff
@@ -85,6 +88,21 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
       )}
+      {showTotpLink && (
+        <Button
+          appearance="link"
+          tabIndex={5}
+          type="button"
+          onClick={() =>
+            (window.location.href = window.location.href.replace(
+              "&use_backup_code=true",
+              "",
```
nsklikas, line 99: question: afaict this causes the page to reload, but we render the exact same flow. Can't we just re-render the page without making the calls to the backend? Would this make sense?

edlerd: yes, good idea. changed the approach a bit to avoid the refetch of the whole page.

ui/components/NodeInputSubmit.tsx
```diff
@@ -77,14 +83,42 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
       )}
+      {showTotpLink && (
+        <Button
+          appearance="link"
+          tabIndex={5}
+          type="button"
+          onClick={() => {
+            const newQuery = { ...router.query };
+            delete newQuery.use_backup_code;
+            void router.push(
```
nsklikas, line 109: Should we use `push` or `replace`?

With `push` this gets pushed in the browser history, this means that when you press back you will be redirected to the previous page.

With `replace`, this does not get pushed in the browser history, which means that if you press back you will go to the actually first page.

i.e. the current behavior is that if you go to the `Verify your identity` page, click on`Use backup code instead` and then click back, you will be redirected to the `Authentication code` page again.

Whereas with `replace` you would go back to the original login page (where you input email and password). 

We could even make it so that you go back to the application page instead. (cc @lukasSerelis what do you think about this?)

nsklikas (commented):
lgtm, I only have a UX comment which I'm not sure about

ui/components/NodeInputSubmit.tsx
edlerd: As a user I'd expect the extra history entry. So going back brings me the state before clicking any button or link. So I think push is the right thing to use here, not replace. I think replace should be used on automatic redirects, that don't require a user interaction.

lukasSerelis: +1 on what @edlerd said. The only weirdness that could occur here is that if a user, for whatever unexplainable reason hits "I want to use a Backup code" and then  "I want to use Authenticator app" back and forth for a a lot of times, it'd be great if this wouldn't create a new history entry for every time they go back and forth. Not sure if this is possible, or how such implementation would look like. 

This is a very much nice to have cause this isn't a real case that people would do, just that having this sorted would make it into a very polished redirection experience.

nsklikas (dismissed):
> +1 on what @edlerd said. The only weirdness that could occur here is that if a user, for whatever unexplainable reason hits "I want to use a Backup code" and then "I want to use Authenticator app" back and forth for a a lot of times, it'd be great if this wouldn't create a new history entry for every time they go back and forth. Not sure if this is possible, or how such implementation would look like.
> 
> This is a very much nice to have cause this isn't a real case that people would do, just that having this sorted would make it into a very polished redirection experience.

I am not sure how easy it is to implement this, this is what I did and came up with this problem. I agree that this is an edge case. It is not straight forward how this would behave ( when do you replace the history and when do you push? the 1st time? the 2nd?).

BTW, I checked what happens if I try to click back on my google account when I am asked to provide 2FA. If I click back, I get back to the enter email page and I have to do the whole authn flow from the beginning. If I then click forward, I get to the 2FA that I had selected. I have not checked what other providers do and I am not asking to implement it as google, but I guess the idea behind their approach (which seems similar to what I saw in github) is that back means "take me back to the beginning of the flow".

Anyway, I will leave it up to you @edlerd if you wish to make any more changes.

edlerd:
Let's address one thing at a time. The backward/forward thing is best solved later -- already discussing solutions in #279
So I suggest merging this as is to solve one thing and then focus on the next :)

nsklikas (approved):

shipperizer (dismissed):

edlerd:
rebased on main, can this be merged?

BarcoMasile (dismissed):

nsklikas (approved):

ui/next.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  basePath: '/ui',
```
nsklikas, line 2: Won't this break when running behind a proxy that serves the app under a path?

BarcoMasile: I think so, we should probably have an issue to address the dynamic base path same way we (will) do in Admin UI (future because it's still not merged).

edlerd: Yes, and we don't need the base path config here after all. Removed it just now.

BarcoMasile (approved):

nsklikas (approved):

PR #298, opened 2024-09-30
fix(request) avoid using request url in the ui. fixes #289

### LINKED ISSUE #289: Invalid checks/redirects in UI
Not sure if this is what causes the bug or not, but in mulitple places in the frontend we are checking if the response from the backend contains a `request_url` (and sometimes we even redirect the user there), eg try to go to https://iam.dev.canonical.com/stg-identity-jaas-dev-login-ui/ui/reset_password. I thought we caught those on review, but it looks like some of these changes went through.

From a quick search:
- https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/ui/pages/reset_password.tsx#L54
- https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/ui/pages/reset_email.tsx#L40
- https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/ui/pages/setup_secure.tsx#L50
- https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/ui/pages/setup_passkey.tsx#L46
- https://github.com/canonical/identity-platform-login-ui/blob/2357551f634c9be6986260366f1a5f391ae6f210/ui/pages/setup_backup_codes.tsx#L46

I am pretty sure that this is not correct, but I am not sure what these checks are trying to accomplish. In a production environment all these are pointing to admin APIs, which are not exposed to the public internet. The `request_url` is the URL that the backend used to call Kratos, there is no reason to call the same URL from the frontend.

_Originally posted by @nsklikas in https://github.com/canonical/identity-platform-login-ui/issues/247#issuecomment-2345600221_

edlerd:
# Done

- fix(request) avoid using request url in the ui.

fixes #289

### DIFF
--- a/ui/pages/setup_secure.tsx
+++ b/ui/pages/setup_secure.tsx
@@ -48,19 +48,15 @@ const SetupSecure: NextPage = () => {
           ? returnTo.toString()
           : window.location.pathname.replace("setup_secure", "setup_complete"),
       })
-      .then(async ({ data }) => {
-        const pwParam = pwChanged ? pwChanged.toString() : "false";
-        router.query.flow = data.id;
-        router.query.pw_changed = pwParam;
-
-        await router.replace(
-          {
-            pathname: window.location.pathname,
-            query: router.query,
-          },
-          undefined,
-          { shallow: true },
-        );
+      .then(({ data }) => {
+        const pwParam = pwChanged ? `&pw_changed=${pwChanged.toString()}` : "";
+        if (flowId !== data.id) {
+          window.history.replaceState(
+            null,
+            "",
+            `./reset_password?flow=${data.id}${pwParam}`,
+          );
+        }
         setFlow(data);
       })
       .catch(handleFlowError("settings", setFlow))
--- a/ui/pages/reset_email.tsx
+++ b/ui/pages/reset_email.tsx
@@ -36,17 +36,13 @@ const ResetEmail: NextPage = () => {
       .createBrowserRecoveryFlow({
         returnTo: returnTo ? String(returnTo) : undefined,
       })
-      .then(async ({ data }) => {
-        if (data.request_url !== undefined) {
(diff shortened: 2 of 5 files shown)


nsklikas (commented):
I thought setting the `flow=<flow_id>` in the url was intentional, are we sure we want to remove that?

edlerd:
> I thought setting the `flow=<flow_id>` in the url was intentional, are we sure we want to remove that?

Yes, you are right. Changed the approach and unified a bit across all pages.

ui/pages/setup_secure.tsx
```diff
@@ -48,19 +48,15 @@ const SetupSecure: NextPage = () => {
...
-            pathname: window.location.pathname,
-            query: router.query,
-          },
-          undefined,
-          { shallow: true },
-        );
+      .then(({ data }) => {
+        const pwParam = pwChanged ? `&pw_changed=${pwChanged.toString()}` : "";
+        if (flowId !== data.id) {
+          window.history.replaceState(
```
nsklikas, line 54: question: Is `router.replace` and `window.history.replaceState` the same API and the 1st is the nextjs way of interacting with the history API? In that case why don't we use the tool provided from the framework?

nsklikas (approved):
lgtm, but I have a question

ui/pages/setup_secure.tsx
edlerd: Basically: yes. The problem with next router is we can't use relative URLs. I don't see any advantage over using the router that using the vanilla api.

nsklikas: Cool, thanks

PR #299, opened 2024-10-02
feat: add security headers

BarcoMasile:
## Description
Add missing security headers to UI related responses.

Fixes #297

PR #301, opened 2024-10-03
Sign in with security key not working if email is cached

nsklikas:
IAM-1114

Based on https://github.com/canonical/identity-platform-login-ui/pull/292

I made some changes to how the login flows are created in the backend, @edlerd (and @natalian98) please have a look and try it out. If this is fine, @edlerd feel free to either merge this to your branch or just push any necessary changes to this branch.

Cases that should be tested for this PR:
- First time login/authenticator setop
- Inserting username/password, going back from totp page, going forward and inserting code
- Inserting username/password, going back from totp page, inserting username/password again, inserting code
- Inserting username/password, going back from totp page, logging in with github

All the above should result in the user being redirected back to the application

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -169,14 +169,14 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 	q := r.URL.Query()
 	flowId := q.Get("flow")
 
-	body, err := a.service.ParseLoginFlowMethodBody(r)
+	body, cookies, err := a.service.ParseLoginFlowMethodBody(r)
 	if err != nil {
 		a.logger.Errorf("Error when parsing request body: %v\n", err)
 		http.Error(w, "Failed to parse login flow", http.StatusInternalServerError)
 		return
 	}
 
-	loginFlow, cookies, err := a.service.GetLoginFlow(r.Context(), flowId, r.Cookies())
+	loginFlow, _, err := a.service.GetLoginFlow(r.Context(), flowId, r.Cookies())
 	if err != nil {
 		a.logger.Errorf("Error when getting login flow: %v\n", err)
 		http.Error(w, "Failed to get login flow", http.StatusInternalServerError)
@@ -195,7 +195,7 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	flow, cookies, err := a.service.UpdateLoginFlow(r.Context(), flowId, *body, r.Cookies())
+	flow, cookies, err := a.service.UpdateLoginFlow(r.Context(), flowId, *body, cookies)
 	if err != nil {
 		a.logger.Errorf("Error when updating login flow: %v\n", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -29,6 +29,11 @@ type API struct {
 	logger logging.LoggerInterface
 }
 
+type queryParam struct {
+	name  string
+	value string
+}
+
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Post("/api/kratos/self-service/login", a.handleUpdateFlow)
 	mux.Get("/api/kratos/self-service/login/browser", a.handleCreateFlow)
@@ -57,7 +62,7 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	returnTo := q.Get("return_to")
 	aal := q.Get("aal")
 	refresh, err := strconv.ParseBool(q.Get("refresh"))
-	if err == nil {
+	if err != nil {
 		refresh = false
 	}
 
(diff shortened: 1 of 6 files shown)


nsklikas:
Added a fix for the oidc method, I hope the comment explaining why is somewhat clear. I don't feel great having to tweak the kratos cookies that much, but I don't think there is a anything better we can do if we want to have control over the flow.

One thing that I notice that does not happen on staging is that if I go back and forward fast many times, eventually I will just get redirected to the error page. I am not sure if I introduced an error or if this error was introduced by https://github.com/canonical/identity-platform-login-ui/pull/299. I think that the problem is that if we leave the page but some request has not finished, it will not be canceled and it may cause an error. Maybe we should add a controller for cancelling any pending requests on page change. @edlerd @BarcoMasile, any idea?

pkg/kratos/handlers.go
```diff
@@ -582,15 +603,33 @@ func (a *API) deleteKratosSession(w http.ResponseWriter) {
 	// This is hacky as it does not call the Kratos API and is likely to break on
 	// a new Kratos version, but there is no easy way to delete the session
 	// from the Kratos API
-	c := &http.Cookie{
+	c := deleteKratosCookie()
+	http.SetCookie(w, c)
+}
+
+func deleteKratosCookie() *http.Cookie {
```
BarcoMasile, line 610: suggestion(non-blocking): Naming here is a little strange.
May I suggest refactoring this method to `kratosDeletionCookie` or something that reflects the fact that a cookie is returned? I'm saying this because looking at line 366 in service.go it looks misleading/counterintuitive

pkg/kratos/service.go
```diff
@@ -355,12 +354,24 @@ func (s *Service) UpdateLoginFlow(
...
+	c := resp.Cookies()
+	if body.UpdateLoginFlowWithOidcMethod != nil {
+		// If this is an oidc flow, we need to delete the session cookie
+		// A session cookie (probably) means that the user used 1fa, but went back from the 2nd factor screen
+		// If the session cookie is set, then Kratos will redirect the user to the default return to URL
+		// The only way to avoid this is by setting refresh=true, but the user has no kratos session.
+		// This is probably a bug on kratos side, as they check if a session exists to set refresh=false.
+		// But in oidc they only check if the session cookie exists, which is not sufficient as the user may not have
+		// enough aal
+		c = append(c, deleteKratosCookie())
```
BarcoMasile, line 366: question: so what you're saying is that we need to do this here because kratos doesn't handle the refresh=true situation correctly, right?

The surprising thing for me is the check on the `aal`, I was convinced they did check for it when working on mfa enforcement

BarcoMasile: Let's talk about this, my "feeling" is that it should not be necessary, but I may be day-dreaming

BarcoMasile:
> Added a fix for the oidc method, I hope the comment explaining why is somewhat clear. I don't feel great having to tweak the kratos cookies that much, but I don't think there is a anything better we can do if we want to have control over the flow.
> 
> One thing that I notice that does not happen on staging is that if I go back and forward fast many times, eventually I will just get redirected to the error page. I am not sure if I introduced an error or if this error was introduced by #299. I think that the problem is that if we leave the page but some request has not finished, it will not be canceled and it may cause an error. Maybe we should add a controller for cancelling any pending requests on page change. @edlerd @BarcoMasile, any idea?

@nsklikas @edlerd 
So #299 just adds security headers which put a constraint on what the browser can trust as a source, so that should not have introduced that behavior.
I think that when moving back and forth fast several times fast, many things could cause issues.
If the "use case" is not realistic I think we can move on and disregard that.
If, instead, it happens also for "regular" user behaviors then we need to dig.

About canceling request it could be a solution. Probably a simpler solution could be to "buffer" requests on the backend (or even the frontend) so we don't overcrowd Kratos with too many requests.
I hope we don't actually need this but if "regular" user behavior requests it, we can look at it.

nsklikas:
> > Added a fix for the oidc method, I hope the comment explaining why is somewhat clear. I don't feel great having to tweak the kratos cookies that much, but I don't think there is a anything better we can do if we want to have control over the flow.
> > One thing that I notice that does not happen on staging is that if I go back and forward fast many times, eventually I will just get redirected to the error page. I am not sure if I introduced an error or if this error was introduced by #299. I think that the problem is that if we leave the page but some request has not finished, it will not be canceled and it may cause an error. Maybe we should add a controller for cancelling any pending requests on page change. @edlerd @BarcoMasile, any idea?
> 
> @nsklikas @edlerd So #299 just adds security headers which put a constraint on what the browser can trust as a source, so that should not have introduced that behavior. I think that when moving back and forth fast several times fast, many things could cause issues. If the "use case" is not realistic I think we can move on and disregard that. If, instead, it happens also for "regular" user behaviors then we need to dig.
> 
> About canceling request it could be a solution. Probably a simpler solution could be to "buffer" requests on the backend (or even the frontend) so we don't overcrowd Kratos with too many requests. I hope we don't actually need this but if "regular" user behavior requests it, we can look at it.

That PR disables the bfcache, which IMO causes this behavior. On staging when you go back most the calls are not re-done, but on this branch going back or forward triggers the request again.

My opinion is that even though things like this may seem like edge cases, on a real world scenario where you may have weird traffic spikes, stupid users, outages, etc they will happen more often than you expect. 

I tested it on `main` and I am pretty sure that disabling the cache caused this behavior, so it should be unrelated to this PR. Maybe we should consider having a better look on that header.

pkg/kratos/service.go
nsklikas: see https://github.com/ory/kratos/blob/master/selfservice/strategy/oidc/strategy.go#L344-L356 vs https://github.com/ory/kratos/blob/master/selfservice/flow/login/handler.go#L165

let's discuss after the daily

BarcoMasile (approved):

nsklikas:
Had a chat with @edlerd and decided to remove his commit, should be good to go now

BarcoMasile (approved):

edlerd (approved):
LGTM :+1:

PR #309, opened 2024-10-08
Fix backup codes setup

nsklikas:
After the recent changes, setting up the backup codes was broken.

If you clicked on the `generate backup codes` button, nothing would happen. This PR fixes that


shipperizer (approved):

natalian98 (approved):
lgtm, tried out and it works well

PR #311, opened 2024-10-08
Cleanup redundant calls to reset the flow in case we redirect right afterwards

### LINKED ISSUE #310: Remove resetting of the flow
In some parts of the frontend we unset the flow (https://github.com/search?q=repo%3Acanonical%2Fidentity-platform-login-ui%20setFlow(undefined)&type=code). I am not sure why that would be needed outside of the error page (even there I am not sure why that's needed). 

We need to investigate this and remove and unnecessary actions. After discussing it with @edlerd, we should be able to remove the unsetting when we redirect the user somewhere, but it may be necessary to keep it in other places (maybe there is a backend bug, where the flow is not sent)

edlerd:
# Done
- Cleanup redundant calls to reset the flow in case we redirect right afterwards

fixes #310

### DIFF
--- a/ui/pages/reset_password.tsx
+++ b/ui/pages/reset_password.tsx
@@ -63,7 +63,6 @@ const ResetPassword: NextPage = () => {
       .catch(handleFlowError("settings", setFlow))
       .catch((err: AxiosError<string>) => {
         if (err.response?.data.trim() === "Failed to create settings flow") {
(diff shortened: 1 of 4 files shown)


nsklikas (approved):
lgtm, thanks

shipperizer (approved):

PR #312, opened 2024-10-08
Ensure a completion message is present after setting up totp.

### LINKED ISSUE #302: First time login improved flow
It'd be best from UX point of view to add a confirmation screen before the automatic redirect to the application. So the flow should be something like this:
image
Or in screens it would be 

1. Trying to sign in for the first time 
image

2. Required TOTP setup
image

3. Setup complete, message to confirm that action has been successful (also good place to show any errors if some occur). And automatic redirect within 3 seconds.
image

4. Landing on the 3rd party application page
image

_Originally posted by @lukasSerelis in https://github.com/canonical/identity-platform-login-ui/issues/275#issuecomment-2376117255_

edlerd:
# Done
- Ensure a completion message is present after setting up totp
- Automatically redirect back to the application
- Fix wrong page URL replacement, when arriving at the setup secure page.

fixes #302

### DIFF
--- a/ui/pages/setup_secure.tsx
+++ b/ui/pages/setup_secure.tsx
@@ -54,13 +54,17 @@ const SetupSecure: NextPage = () => {
           window.history.replaceState(
             null,
             "",
-            `./reset_password?flow=${data.id}${pwParam}`,
+            `./setup_secure?flow=${data.id}${pwParam}`,
           );
         }
         setFlow(data);
       })
       .catch(handleFlowError("settings", setFlow))
       .catch(async (err: AxiosError<string>) => {
+        if (err.response?.data.trim() === "Failed to create settings flow") {
+          window.location.href = "./login";
+          return;
+        }
         return Promise.reject(err);
       });
   }, [flowId, router, router.isReady, returnTo, flow]);
@@ -80,15 +84,13 @@ const SetupSecure: NextPage = () => {
           },
         })
         .then(({ data }) => {
-          if ("redirect_to" in data) {
-            window.location.href = data.redirect_to as string;
-            return;
+          if (data?.return_to && !data.return_to.endsWith("/setup_complete")) {
+            // we do a have a valid return_to, show the completion step, and it will redirect after 3 seconds
+            window.location.href = `./setup_complete?flow=${data.id}}`;
+          } else {
+            // no return_to, show the setup complete page, and it will ask the user to close the tab
+            window.location.href = "./setup_complete";
           }
-          if (data?.return_to) {
-            window.location.href = data?.return_to;
-            return;
-          }
-          window.location.href = "./error";
         })
         .catch(handleFlowError("settings", setFlow));
     },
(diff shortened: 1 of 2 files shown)


ui/pages/setup_secure.tsx
```diff
@@ -81,11 +81,11 @@ const SetupSecure: NextPage = () => {
         })
         .then(({ data }) => {
           if ("redirect_to" in data) {
-            window.location.href = data.redirect_to as string;
+            window.location.href = `./setup_complete?flow=${data.id}`;
             return;
           }
```
nsklikas, line 86: This is not needed, we can remove this case. The backend should always return the flow

nsklikas (commented):
LGTM, only a small comment

ui/pages/setup_secure.tsx
edlerd: removed this bit.

ui/pages/setup_secure.tsx
```diff
@@ -80,12 +80,11 @@ const SetupSecure: NextPage = () => {
...
-          }
           if (data?.return_to) {
-            window.location.href = data?.return_to;
+            const flowParam = data.return_to.endsWith("/setup_complete")
+              ? ""
+              : `?flow=${data.id}`;
+            window.location.href = `./setup_complete${flowParam}`;
             return;
           }
           window.location.href = "./error";
```
nsklikas, line 90: I would remove this condition, let the other page handle if there is no return_to. If you want to be safe, we can check the state of the flow (but I'm not sure if that's needed)

IMO Redirecting to the error page does not make sense

ui/pages/setup_secure.tsx
```diff
@@ -80,12 +80,11 @@ const SetupSecure: NextPage = () => {
...
-          if ("redirect_to" in data) {
-            window.location.href = data.redirect_to as string;
-            return;
-          }
           if (data?.return_to) {
-            window.location.href = data?.return_to;
+            const flowParam = data.return_to.endsWith("/setup_complete")
+              ? ""
+              : `?flow=${data.id}`;
+            window.location.href = `./setup_complete${flowParam}`;
```
nsklikas, line 87: What if the return_to page is not `setup_complete` but something else? Shouldn't this be:
```suggestion
            window.location.href = `./${data.return_to}${flowParam}`;
```
?

ui/pages/setup_secure.tsx
nsklikas, line 90: I don't think we should redirect to the error page in this case, let's just send them to the `setup_complete` page with the flow and let that page handle it.

nsklikas (approved):

nsklikas (approved):

ui/pages/setup_secure.tsx
nsklikas: you are right

shipperizer (approved):

Issue #316, opened 2024-10-09
Overwritten `return_to` url when MFA setup must be enforced before backup codes check


natalian98:
If the user has 3 or less backup codes left, (s)he should be presented with a “Regenerate backup codes” screen and offered to regenerate the codes or continue to sign in:
![image](https://github.com/user-attachments/assets/eb7efa56-eb1f-4704-825f-002d31ffbb82)

However, if the user has no totp authenticator linked, in the above case the user will be first redirected to set it up (see [here](https://github.com/canonical/identity-platform-login-ui/blob/dd604ec001b4be1bc826bbb59fa75cd4970968bb/pkg/kratos/handlers.go#L238)) and after doing so, redirected to `/ui/setup_complete` without ever going to the backup codes screen.

It is a minor bug but may require admin intervention if the user doesn't end up adding totp and all the backup codes are eventually used.

I think the issue is both in frontend and backend:
- if MFA setup is enforced, we redirect to `setup_secure` and return without getting to [this](https://github.com/canonical/identity-platform-login-ui/blob/dd604ec001b4be1bc826bbb59fa75cd4970968bb/pkg/kratos/handlers.go#L242) check.
- the latter `return_to` is overwritten [here](https://github.com/canonical/identity-platform-login-ui/blob/dd604ec001b4be1bc826bbb59fa75cd4970968bb/ui/pages/setup_secure.tsx#L89) and is always forcing the redirection to `setup_complete` page.

nsklikas:
The `setup_complete` page is supposed to fetch the `return_to` from the flow and redirect the user there after a short period of time. I think that the issue was always there and the backend should handle it, although it is not clear to me right now how that should be done (maybe we could over-ride the return_to of the login flow with one to regenerate the codes and then return the user to the original return_to)

BarcoMasile:
> * if MFA setup is enforced, we redirect to `setup_secure` and return without getting to [this](https://github.com/canonical/identity-platform-login-ui/blob/dd604ec001b4be1bc826bbb59fa75cd4970968bb/pkg/kratos/handlers.go#L242) check.
> * the latter `return_to` is overwritten [here](https://github.com/canonical/identity-platform-login-ui/blob/dd604ec001b4be1bc826bbb59fa75cd4970968bb/ui/pages/setup_secure.tsx#L89) and is always forcing the redirection to `setup_complete` page.

What if we perform the backup code check before the mfa one?
Could that help?
@natalian98

[...]

PR #317, opened 2024-10-10
refactor: move backup codes check before mfa setup enforcement check

natalian98:
fixes https://github.com/canonical/identity-platform-login-ui/issues/316

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -234,13 +234,13 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 
 	setCookies(w, cookies)
 
-	if shouldEnforceMfa {
-		a.mfaSettingsRedirect(w, *loginFlow.ReturnTo)
+	if shouldRegenerateBackupCodes {
+		a.lookupSecretsSettingsRedirect(w, flowId)
 		return
 	}
 
-	if shouldRegenerateBackupCodes {
-		a.lookupSecretsSettingsRedirect(w, flowId)
+	if shouldEnforceMfa {
+		a.mfaSettingsRedirect(w, *loginFlow.ReturnTo)
 		return
 	}
 


nsklikas (commented):
question: doesn't this run into the reverse problem? (i.e. authenticator setup won't be enforced?)

natalian98:
> question: doesn't this run into the reverse problem? (i.e. authenticator setup won't be enforced?)

If the user continues to sign in and a totp authenticator is not linked, he will still be enforced to set it up.

shipperizer (commented):
don't we have any test failing due to this change?

nsklikas:
I am having trouble testing this locally, as for some reason using backup codes is very flaky in firefox (the wrong method is sent most of the time). BUT if I understand correctly, we don't care about checking if the authenticator is setup because the user will be forced to set it up on their next login?

natalian98:
> I am having trouble testing this locally, as for some reason using backup codes is very flaky in firefox (the wrong method is sent most of the time).

I tested on chromium and chrome and didn't experience that at all, will double check on firefox.

> BUT if I understand correctly, we don't care about checking if the authenticator is setup because the user will be forced to set it up on their next login?

That's right, the totp check repeats on every login

nsklikas:
Same issue happens on chromium

natalian98:
> Same issue happens on chromium

are you sure you click on "use backup code" first if that's not the only option?

nsklikas:
Yeah and it also happens when backup_codes is the only options.

BTW: How can you test it if backup_codes is not the only option?

nsklikas:
Ok, I managed to test it. One problem that I faced is that after get the new backup codes, get redirected to setup the authenticator and then I am not redirected back to the application, but instead I get to the setup_complete page and stay there

natalian98:
That was always the case and independent of the backup codes. If you don't have any 2fa set up, after linking totp app I also end up at `setup_complete` page and don't get redirected anywhere

nsklikas:
I realize that, ok let's open an issue about it.

nsklikas (approved):

natalian98:
Looks that https://github.com/canonical/identity-platform-login-ui/pull/312/files attempted to solve it but we're still missing sth.

BarcoMasile (approved):

PR #321, opened 2024-10-10
fix first time login redirect

### LINKED ISSUE #319: No redirection from `setup_complete` page to the application
If you don't have any 2fa set up, after linking totp app on first login you'll end up at `setup_complete` page (`http://localhost:4455/ui/setup_complete?flow=df688f30-98d1-48f6-9a4b-4496f7911152}`) and won't get redirected back to the original url (e.g. grafana).
image

edlerd:
# Done
- fix first time redirect after setting up totp

fixes #319

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"path"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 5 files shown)


natalian98:
Thanks a lot for the fix.

The redirection after setting up totp works well now, but the backup codes case is still a bit broken:
1. When you log in with a backup code and get redirected to `regenerate_backup_codes` screen
2. Then click on "Generate new codes"
3. Create, save and confirm a new set
4. You get redirected to `setup_complete`, but without further going back to the grafana url:
image

The `return_to` parameter was missing in the backend when redirecting to regenerate screen, so I pushed a commit to add it. Could you adjust frontend to behave in the same way as after completing `setup_secure`?

To make testing easier you can switch [this](https://github.com/canonical/identity-platform-login-ui/blob/2029a4dde0ad863db111479c50a381a1d4889ed2/pkg/kratos/service.go#L37) to 12, so that you always get asked if you want to regenerate codes if you used one

edlerd:
Added the forward of the return for the backend code regeneration also to the frontend.

nsklikas (approved):

natalian98 (approved):
works well, thank you

PR #325, opened 2024-10-11
fix ui for totp setup screen, add separators

### LINKED ISSUE #324: TOTP setup UI
Current UI isn't ideal due to mental load presented. There is at the same time QR code, text code and entry field. Ideally we'd either have more seperation or hide the secret behind another screen.

Scenario 1, hiding one-time code secret thing behind a link:
1) Default screen.
image
https://www.figma.com/design/zstuCMdHsKx8Wr9ERZx9Kw/Login-updates-(passkey%2C-MFA%2C-org-links)?node-id=166-8979&node-type=frame&t=GzhBGn7sxcZeBrJj-11

2) If you click can't scan code you see this. Clicking use QR instead takes you to default screen
image
https://www.figma.com/design/zstuCMdHsKx8Wr9ERZx9Kw/Login-updates-(passkey%2C-MFA%2C-org-links)?node-id=235-5519&node-type=frame&t=GzhBGn7sxcZeBrJj-11

Scenario 2:
Show the one-time code secret but better separate them better
image
https://www.figma.com/design/zstuCMdHsKx8Wr9ERZx9Kw/Login-updates-(passkey%2C-MFA%2C-org-links)?node-id=235-5730&node-type=frame&t=GzhBGn7sxcZeBrJj-11

edlerd:
# Done
- fix ui for totp setup screen, add separators

fixes #324

### DIFF
--- a/ui/components/NodeText.tsx
+++ b/ui/components/NodeText.tsx
@@ -86,12 +86,23 @@ const Content: FC<Props> = ({ attributes }) => {
 };
 
 export const NodeText: FC<Props> = ({ node, attributes }) => {
+  const isTotpSetup = attributes.id === "totp_secret_key";
+
   return (
     <>
+      {isTotpSetup && <hr />}
       <p data-testid={`node/text/${attributes.id}/label`}>
-        {node.meta.label?.text}
+        {isTotpSetup ? (
+          <>
+            Or <strong>if you can not scan the QR code</strong>, use the
+            provided one-time code
+          </>
+        ) : (
+          node.meta.label?.text
+        )}
       </p>
       <Content node={node} attributes={attributes} />
+      {isTotpSetup && <hr />}
     </>
   );
 };
--- a/ui/components/NodeImage.tsx
+++ b/ui/components/NodeImage.tsx
@@ -9,10 +9,15 @@ interface Props {
 
 export const NodeImage: FC<Props> = ({ node, attributes }) => {
   return (
-    <Image
-      data-testid={`node/image/${attributes.id}`}
-      src={attributes.src}
-      alt={node.meta.label?.text ?? ""}
-    />
+    <>
+      {attributes.id === "totp_qr" && (
+        <p>Open your authenticator app and scan the QR code</p>
+      )}
+      <Image
+        data-testid={`node/image/${attributes.id}`}
+        src={attributes.src}
+        alt={node.meta.label?.text ?? ""}
+      />
+    </>
   );
 };


nsklikas (approved):

natalian98 (approved):

natalian98 (changes requested):
Could you also change the wording to address Lukas' comment?
We have:
image
We'd like to have:
image

edlerd:
> Could you also change the wording to address Lukas' comment?

I assume this is kratos configuration? Because the text is coming from the UI nodes of the kratos flow.

natalian98:
> Can you configure this in Kratos @natalian98 ? As the UI is not in control of the copy.

This is not configurable in kratos, but I assumed you can change the ui nodes values in the frontend. If that's not possible, let's leave it in the current state.

edlerd:
Amended the PR to handle also the special copy.

ui/components/NodeText.tsx
```diff
@@ -86,12 +86,25 @@ const Content: FC<Props> = ({ attributes }) => {
...
+
   return (
     <>
+      {isTotpSetup && <hr />}
       <p data-testid={`node/text/${attributes.id}/label`}>
-        {node.meta.label?.text}
+        {isTotpSetup ? (
+          <>
+            Or <strong>if you can not scan the QR code</strong>, use the
+            provided on-time code
```
natalian98, line 100:
```suggestion
            provided one-time code
```

nsklikas (commented):
This looks fine, but I have some concerns with this approach. It feels like the mapping of custom texts to our own logic is bound to get our of hand. I think that we should:
- Invest some time to evaluate if we can have a more generalized approach to mapping so that the mapping logic is centralized somewhere
- Instead of mapping text to text, we should map attribute_id to text. For example: 
Change `const isTotpQrCode = node.meta.label?.text === "Authenticator app QR code";` to `const isTotpQrCode = node.attributes.id === totp_qr` and 
```
  const isTotpSetup =
    node.meta.label?.text ===
    "This is your authenticator app secret. Use it if you can not scan the QR code.";
``` 
to `const isTotpSetup = node.attributes.id === "totp_secret_key";`
The problem with using the text as an id is that on update the text might change, but IDs are way less likely to change

ui/components/NodeText.tsx
edlerd: whoopsie.

edlerd:
> This looks fine, but I have some concerns with this approach. It feels like the mapping of custom texts to our own logic is bound to get our of hand. I think that we should:

Totally agree, wasn't aware of those stable text ids, which are much better to target. Changed the approach.

nsklikas (approved):

nsklikas (approved):

natalian98 (approved):

PR #328, opened 2024-10-14
improve ui for reset password

### LINKED ISSUE #322: Reset password flow not working and UI improvements
When trying to reset password it doesn't complete the flow. Occasionally also the text-field is showing an error in field which is incorrect.
image

On UI side, back button needs to be there, and Submit button needs to say "Reset password". Error message says "Invalid email address" should say "Incorrect email address"
image

If possible, when we can actually get to the success screen, it'd be great to validate which email address was submitted.
image

edlerd:
# Done
- add the back button (it should just redirect to the login url)
- rename Submit to Reset password
- make sure the field doesn't show the error message once you click on the button

fixes #322

Skipped the "refactor the "Check your email" screen to contain the email address typed in the previous step" as per #322. 
 Because 1. we don't know the email in the 2nd step and 2. the email does not contain any url, only the code to enter. So we need to keep the step with the input box.

### DIFF
--- a/ui/components/NodeInputSubmit.tsx
+++ b/ui/components/NodeInputSubmit.tsx
@@ -42,9 +42,21 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
   const showTotpLink =
     (node.meta.label as unknown as { hasTotpLink: boolean })?.hasTotpLink ??
     false;
+  const showBackLink = node.meta?.label?.text === "Reset password";
 
   return (
     <>
+      {showBackLink && (
+        <Button
+          tabIndex={3}
+          type="button"
+          onClick={() => {
+            void window.history.back();
+          }}
+        >
+          Back
+        </Button>
+      )}
       <Button
         appearance={
           node.group === "password" ||
--- a/ui/components/NodeInputEmail.tsx
+++ b/ui/components/NodeInputEmail.tsx
@@ -14,17 +14,26 @@ export const NodeInputEmail: FC<NodeInputProps> = ({
   const [hasLocalValidation, setLocalValidation] = React.useState(false);
 
   const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-  const isInvalid = !emailRegex.test((value as string) ?? "");
+  const isInvalid = !emailRegex.test((value as string) ?? "") && value !== "";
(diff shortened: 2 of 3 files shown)


ui/components/NodeInputSubmit.tsx
```diff
@@ -42,9 +42,21 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
+  const showBackLink = node.meta?.label?.text === "Reset password";
 
   return (
     <>
+      {showBackLink && (
+        <Button
+          tabIndex={3}
+          type="button"
+          onClick={() => {
+            void window.history.back();
```
nsklikas, line 54: IMHO back buttons are a pain both from a dev and a UX perspective.

question: what happens if I access this page directly? My guess is that I will be redirected back to wherever I was, @lukasSerelis is that we want or do we always want to be redirected to the login page? What happens if I am logged in and get redirected to the login page? IMO if we want to add back buttons, we should hard code in the UI the succession of pages in the flow and redirect the user there based on these hardcoded values, not based on the browser's history stack.

edlerd: The problem with hardcoding is, that we lose the challenge from the app. So when a user came from grafana, they won't log in to grafana any more. Using history was the best of all worlds I could think of. Technically, this is always the correct behaviour.

nsklikas: Oh, that's right. I didn't think of this.

Then I am not sold on whether we need this back button, but I will let it be :P

edlerd: Happy to remove it, as that simplifies the code. @lukasSerelis wdyt?

lukasSerelis: Even if this is just a call for history back I would like to keep the back button. Login flows are notoriously finicky with going back (similar to checkout flows), so being explicit with back actions here is helpful. Otherwise, it's not clear at all what are the options if I am not sure I want to reset anymore.

shipperizer:
@edlerd commit message doesn't follow conventional commits, therefore our machinery doesn't trigger a new release, can u use the `fix:` prefix please?

natalian98 (approved):

PR #330, opened 2024-10-14
Return error if no login_challenge or return_to

### LINKED ISSUE #329: `login_challenge` not created when user attempts to sign in directly in login ui
If you go directly to the login page `https://iam.dev.canonical.com/stg-identity-jaas-dev-login-ui/ui/login` without going through a hydra client first (e.g. grafana), there is no login challenge created.
This causes any login attempt to fail: after providing password and totp code, login ui displays an error:
image

```
https://iam.dev.canonical.com/stg-identity-jaas-dev-login-ui/api/kratos/self-service/errors?id=undefined
"Error when getting flow error: 404 Not Found"
```
That is because the call to `/self-service/login/browser` should include the `login_challenge` parameter.

We should fix the flow logic for when the user attempts to log in directly in login ui to enable users to manage their account settings.
Note: this error can also be reproduced in a local dev setup, so recent changes haven't fixed it.

nsklikas:
IAM-1152

With this change, if you try to access localhost:4455/ui/login, the backend will return an error (400) and the browser will be redirected to the error page.

Closes #329

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -66,6 +66,10 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if returnTo == "" {
+		if loginChallenge == "" {
+			http.Error(w, "One of return_to or login_challenge must be provided", http.StatusBadRequest)
+			return
+		}
 		returnTo, err = a.returnToUrl(loginChallenge)
 		if err != nil {
 			// this should never happen if app is properly configured
--- a/pkg/kratos/handlers_test.go
+++ b/pkg/kratos/handlers_test.go
@@ -35,6 +35,32 @@ const (
 //go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_logger.go -source=../../internal/logging/interfaces.go
 //go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_kratos.go -source=./interfaces.go
 
+func TestHandleCreateFlowWithoutParams(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockLogger := NewMockLoggerInterface(ctrl)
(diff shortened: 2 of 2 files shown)


pik4ez-canonical (approved):

nsklikas:
There is an argument to be made about whether this change is correct or not. Currently (on `main`) if the user logs in, they will be shown an error page, BUT they can go to the login page and then choose to reset password. BUT with this change if the user tries to go to the login page, they will be redirected to the error page directly. While this is better UX in the case where the user tries to login (because they will be shown the error earlier), it stops the user from clicking the `reset password` button. To use that button, the user needs to come with a return_to, a login_challenge or they need to access the reset_password page directly. Probably with the current state of the login UI (where we don't have a home page) this is acceptable, but I would like to hear what other people think.

BarcoMasile:
> There is an argument to be made about whether this change is correct or not. [...] Probably with the current state of the login UI (where we don't have a home page) this is acceptable, but I would like to hear what other people think.

I wrote a comment [here](https://github.com/canonical/identity-platform-login-ui/issues/329#issuecomment-2413232664).
To add on this: I think it's ok to only let users reset password when actually trying to log in to a 3rd party app (like Grafana), this is what usually happens with most providers, which have a separate section/app/page to handle account details and account operations.

We should aim towards having an account page accessible always on login ui, but that needs design, APIs probably etc..
So for now, I think we can go on with this if we're sure we're not "blocking" regular calls to login ui (which I don't think is the case).

BarcoMasile (approved):

natalian98 (approved):
An easy approach could be to add a `return_to` if none is set, for example `http://localhost:4455/ui/login?return_to=http://localhost:4455/ui/reset_password`. That would allow users to self-manage their accounts. That said, it's proabably not the best UX to suggest that password should be changed every time the user logs in this specific way.

I'm fine to merge this PR and set the `return_to` to a self-service menu once we have one. I'll open an issue to track it.

PR #345, opened 2024-10-22
ci: add oci-factory publishing

shipperizer:


.github/workflows/publish.yaml
```diff
@@ -79,3 +82,35 @@ jobs:
...
+
+  oci-factory:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4
+      - name: Golang setup
+        uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22.0'
```
natalian98, line 94: Does it not conflict with login ui using [v1.21](https://github.com/canonical/identity-platform-login-ui/blob/67e1b767b2d1409f32699da4ec7d2f768fa5f3a5/go.mod#L3)?

natalian98 (approved):

.github/workflows/publish.yaml
shipperizer: nope it is just to install the OCI factory cli, the image gets built based on the rockcraft specs

PR #346, opened 2024-10-23
Check hydra request skip value

### LINKED ISSUE #300: max_age is not respected
Currently if we find that the user already has a session, we always accept the hydra login challenge.

This means that we don't respect the max_age that was requested by the client or hydra. If we used the Kratos createLoginFlow API, then this would be taken care of, as it fetches the hydra login request and checks whether it should `skip` authn (see https://github.com/ory/kratos/blob/master/selfservice/flow/login/handler.go#L526)

There are a couple of issues with letting the kratos createLoginFlow API to handle the login challenge though:
- If this is the first time the user logs in with this client, then they will be forced to log in even if they have an active kratos session
- We won't be able to redirect the user back to the app the first time they login. The reason for this is that the first time a user logs in, they are redirected to the settings page and then they are redirected back to the login page. Even though they will have an active session, they will be forced to re-authenticate because kratos will try to refresh the session (hydra flow.skip will be false)

nsklikas:
IAM-1127

Fetch the hydra login request and check the `Skip` value to decide whether we should accept it if a session exists. Otherwise (if this isn't a hydra flow or if there is no session) we leave it to Kratos. This fixes the issue where `max_age` was not respected.

Closes #300

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -94,6 +94,14 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
+	}
+
+	forceLogin, err := a.service.MustReAuthenticate(r.Context(), loginChallenge, session)
+	if err != nil {
+		a.logger.Errorf("Failed to fetch hydra flow: %v", err)
+		http.Error(w, "Failed to fetch hydra flow", http.StatusInternalServerError)
+	}
+	if !forceLogin {
 		response, cookies, err = a.handleCreateFlowWithSession(r, session, loginChallenge)
 	} else {
 		response, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -2,6 +2,8 @@ package kratos
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -18,12 +20,14 @@ import (
 
 const RegenerateBackupCodesError = "regenerate_backup_codes"
 const KRATOS_SESSION_COOKIE_NAME = "ory_kratos_session"
+const LOGIN_UI_STATE_COOKIE = "login_ui_state"
 
 type API struct {
-	mfaEnabled  bool
-	service     ServiceInterface
-	baseURL     string
-	contextPath string
+	mfaEnabled    bool
+	service       ServiceInterface
+	baseURL       string
+	contextPath   string
+	cookieManager AuthCookieManagerInterface
 
 	logger logging.LoggerInterface
 }
@@ -88,18 +92,25 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			a.logger.Errorf("Failed check for MFA: %v", err)
 			http.Error(w, "Failed check for MFA", http.StatusInternalServerError)
-		}
-		if shouldEnforceMfa {
-			a.mfaSettingsRedirect(w, returnTo)
 			return
 		}
-
+		if shouldEnforceMfa {
+			a.mfaSettingsRedirect(w, returnTo, loginChallenge)
+			return
+		}
 	}
 
-	forceLogin, err := a.service.MustReAuthenticate(r.Context(), loginChallenge, session)
+	c, err := a.cookieManager.GetStateCookie(r)
+	if err != nil {
+		a.logger.Errorf("Failed to parse state cookie: %v", err)
+		http.Error(w, "Failed to parse state cookie", http.StatusInternalServerError)
+		return
+	}
+	forceLogin, err := a.service.MustReAuthenticate(r.Context(), loginChallenge, session, c)
 	if err != nil {
 		a.logger.Errorf("Failed to fetch hydra flow: %v", err)
 		http.Error(w, "Failed to fetch hydra flow", http.StatusInternalServerError)
+		return
 	}
 	if !forceLogin {
 		response, cookies, err = a.handleCreateFlowWithSession(r, session, loginChallenge)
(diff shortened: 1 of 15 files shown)


pkg/kratos/interfaces.go
```diff
@@ -29,6 +29,8 @@ type AuthorizerInterface interface {
 type ServiceInterface interface {
 	CheckSession(context.Context, []*http.Cookie) (*kClient.Session, []*http.Cookie, error)
 	AcceptLoginRequest(context.Context, string, string) (*hClient.OAuth2RedirectTo, []*http.Cookie, error)
+	GetLoginRequest(context.Context, string) (*hClient.OAuth2LoginRequest, []*http.Cookie, error)
```
BarcoMasile, line 32: food for thought: this method seems to only be used in the `MustReAuthenticate`, so maybe we could avoid exposing it?
You already wrote tests for it so I'm happy keeping it here

pkg/kratos/service.go
```diff
@@ -117,6 +117,41 @@ func (s *Service) AcceptLoginRequest(ctx context.Context, identityID string, lc
...
+		Execute()
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return redirectTo, resp.Cookies(), nil
+}
+
+func (s *Service) MustReAuthenticate(ctx context.Context, hydraLoginChallenge string, session *kClient.Session) (bool, error) {
```
BarcoMasile, line 136: suggestion: I would add tracing here also since this method might actually invoke an external API (though the GetLoginRequest (which is already traced)

BarcoMasile (approved):
Great work!
See my microscopic comments, but I think it's LGTM

pkg/kratos/interfaces.go
nsklikas: Makes sense, I will remove it from the interface

pkg/kratos/service.go
nsklikas: Yeah, I wasn't sure if that was needed or not because the `GetLoginRequest` function is doing the external call and that one is already traced, I wasn't sure on which function to start the trace.

You think that both functions should start a trace or only `MustReAuthenticate`?

BarcoMasile (dismissed):

nsklikas:
As @natalian98 pointed out, this PR breaks the first login experience. 

If this is the first time a user logs in to an application, the `Skip` value in hydra will be `False` (the reason for this is that the user has not given consent for this application, so they can't skip login even if they have a session). This means that if this the first time the user logs in, `Skip` will be `False`. So after they setup their authenticator they will be redirected back to the original page, which will call the create login flow backend. Instead of redirecting the user back to hydra, the user will be asked to authenticate again. 

The simplest solution to overcome this is to use @BarcoMasile cookie implementation form the admin UI. This will solve our problem pretty easily, we will just have to add a short lived cookie with a value in the following format:
```json
{"login_challenge": "<login_challenge>", "totp_setup": true}
```
This format should be extensible in case we want to add more information.

This will be a breaking change as we will have to add a cookie secret configuration.

@natalian98 @BarcoMasile should I add this in this PR or should I just add a dummy cookie solution and we refactor it in another PR?

natalian98 (commented):
Could you add `COOKIES_ENCRYPTION_KEY` to the docker-compose file and readme?

I'm having some trouble testing the new changes, going to login page results in below errors and redirects me to the error page:
```
{"severity":"error","@timestamp":"2024-10-25T10:48:04.550376835+02:00","message":"Failed to parse state cookie: http: named cookie not present"}

{"severity":"error","@timestamp":"2024-10-25T10:50:43.077272354+02:00","message":"Error when getting flow error: 404 Not Found\n"}
```

pkg/kratos/handlers.go
```diff
@@ -346,7 +354,9 @@ func (a *API) is40xError(err error) bool {
 	return false
 }
 
-func (a *API) mfaSettingsRedirect(w http.ResponseWriter, returnTo string) {
+func (a *API) mfaSettingsRedirect(w http.ResponseWriter, r *http.Request, returnTo, loginChallenge string) {
```
BarcoMasile, line 357: micro-nitpick: if you only need the context from the request, consider just passing the context as an argument instead of the request pointer

BarcoMasile (dismissed):
Great work!

nsklikas:
I had to switch things a little, but I think it works fine now. I had to set the remember field in the hydra session to make it work correctly.

`max_age` is respected (`0` means force relogin, `-1` means keep session)
1st login seems to work fine as well

@natalian98 Please let me know if the error still happens

natalian98 (dismissed):
Works well now, thank you

pkg/kratos/handlers.go
```diff
@@ -88,12 +92,26 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
...
+	c, err := a.cookieManager.GetStateCookie(r)
+	if err != nil {
+		a.logger.Errorf("Failed to parse state cookie: %v", err)
+		http.Error(w, "Failed to parse state cookie", http.StatusInternalServerError)
+		return
+	}
+	forceLogin, err := a.service.MustReAuthenticate(r.Context(), loginChallenge, session, c)
+	if err != nil {
+		a.logger.Errorf("Failed to fetch hydra flow: %v", err)
+		http.Error(w, "Failed to fetch hydra flow", http.StatusInternalServerError)
```
BarcoMasile, line 112: I think you need a `return` here

pkg/kratos/handlers.go
```diff
@@ -350,10 +370,16 @@ func (a *API) mfaSettingsRedirect(w http.ResponseWriter, returnTo string) {
 
 	errorId := "session_aal2_required"
 
+	if loginChallenge != "" {
+		h := md5.New()
+		h.Write([]byte(loginChallenge))
+		sessionId = base64.StdEncoding.EncodeToString(h.Sum(nil))
```
BarcoMasile, line 376: possible issue: use `base64.URLEncoding`

BarcoMasile (approved):
LGTM but the `return` is actually needed, we could do it later anyway

nsklikas:
@BarcoMasile I think I addressed both your comments, have a look again to see if it's fine

BarcoMasile (approved):

natalian98 (approved):

PR #349, opened 2024-11-05
Add end to end test suite

edlerd:
# Done
- add playwright as e2e framework
- implemented 3 test cases for first time login + totp setup, reset password and to setup backup codes
- added screenshots for all tests
- added documentation on how to run the tests
- schedule tests as part of the ci


ui/tests/login-first-time.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+import { userPassLogin } from "./helpers/login";
+
+test("first time login to grafana", async ({ context, page }) => {
+  resetIdentities();
+  await startGrafanaNewUserFlow(page);
+  await userPassLogin(page);
+  await setupTotp(context, page);
+
+  await expect(page.getByText("Account setup complete")).toBeVisible();
+  await expect(page).toHaveScreenshot({ fullPage: true });
```
BarcoMasile, line 14: opinion: this is interesting `toHaveScreenshot` seems like it could be "brittle", I saw we can update screenshots automatically with Playwright. I'd say let's keep this in, but let's be mindful of this function for future tests

BarcoMasile (approved):
Great work!
I'd say next time let's have more commits if it makes sense to split the work

ui/tests/login-first-time.spec.ts
edlerd: I agree. I was hesitant to include it or not. After considering the option carefully, I think the benefits outweigh not having it:
 - have clear documentation of how the screens look like in the repo.
 - text assertions with `expect(...)` only cover partially the functionality available in a given screen. We had bugs in the past, that were due to exposure of additional functionality in screens, where it should be hidden. Having screenshots seems like the right approach against this class of bugs.
 - easily update the screens when updating 3rd party like vanilla changes layout just so slightly. As you mentioned already.
 - see in a PR when screens change, as that PR should contain the old and the new screenshot

BarcoMasile: It makes perfect sense, thank you for explaining this!

ui/tests/scripts/02-start-ui.sh
```diff
@@ -0,0 +1,21 @@
...
+export KRATOS_PUBLIC_URL="http://localhost:4433"
+export KRATOS_ADMIN_URL="http://localhost:4434"
+export HYDRA_ADMIN_URL="http://localhost:4445"
+export BASE_URL="http://localhost:4455"
+export PORT="4455"
+export TRACING_ENABLED="false"
+export LOG_LEVEL="debug"
+export AUTHORIZATION_ENABLED="false"
+
+go run . serve
```
BarcoMasile, line 21: suggestion: you could split this in

```golang
go build . -o app ./
./app serve
```

but feel free to ignore this

.github/workflows/e2etest.yaml
BarcoMasile, line 1: suggestion: we should probably perform the "installation" steps first, and then actually boot the cluster
So if for any reason installation for some of the tools goes wrong, we don't have things spun up that needs to come down

BarcoMasile (approved):

.github/workflows/e2etest.yaml
edlerd: We install the browser right before executing tests in other repos. I have never seen that step fail, so it shouldn't be a problem. I would keep the order this way, because it gives the backend some additional time to boot, while the browser is being fetched.

ui/playwright.config.ts
```diff
@@ -0,0 +1,38 @@
+import type { PlaywrightTestConfig } from "@playwright/test";
+import { devices } from "@playwright/test";
+
+const config: PlaywrightTestConfig = {
+  testDir: "./tests",
+  /* Maximum time one test can run for. */
+  timeout: 30_000,
```
anusha-c18, line 7: We could also configure the `snapshotPathTemplate` option

ui/tests/reset-password.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  await context.clearCookies({ domain: "localhost" });
+  await startGrafanaNewUserFlow(page);
+
+  await userPassLogin(page);
+  await expect(page.getByText("Incorrect username or password")).toBeVisible();
+  await expect(page).toHaveScreenshot({ fullPage: true, maxDiffPixels: 500 });
+
+  await page.getByLabel("Password").fill(PASSWORD_NEW);
+  await page.getByRole("button", { name: "Sign in", exact: true }).click();
```
anusha-c18, line 46: Can we refactor the `userPassLogin` function to accept the password as a parameter and reuse it here as well?

anusha-c18 (changes requested):
This looks good already!
A few small suggestions

ui/playwright.config.ts
```diff
@@ -0,0 +1,38 @@
...
+  retries: process.env.CI ? 2 : 0,
+  workers: 1,
+  reporter: [["html", { fileName: "index.html" }]],
+  use: {
+    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
+    actionTimeout: 0,
+    baseURL: "http://localhost:2345/login/",
+    ignoreHTTPSErrors: true,
+    video: "retain-on-failure",
+    trace: "on-first-retry",
```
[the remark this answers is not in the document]
edlerd: Good idea. Though the viewport is included in the `devices["Desktop Chrome"]` as per the [playwright docs](https://playwright.dev/docs/emulation#viewport), so we don't need to specify it.

ui/tests/reset-password.spec.ts
edlerd: yes, good idea!

ui/playwright.config.ts
edlerd: yes, agreed!

ui/tests/helpers/login.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, Page } from "@playwright/test";
+
+export const USER_EMAIL = "test@example.com";
+export const USER_PASS = "test";
```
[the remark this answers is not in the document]
edlerd: With the generalized user login function, there is one external use case for this variable to make the distinction between old and new password clear in the reset password test case.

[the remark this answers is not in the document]
edlerd: Oh yes, it is. This is not changing the default, but to point out the difference between the two calls in line 41 and 45:

image

anusha-c18 (dismissed):
Thanks for the changes, LGTM 👍

BarcoMasile (dismissed):

ui/tests/scripts/03-start-grafana.sh
nsklikas, line 1: Not sure why we need to use grafana for this, we could use the hydra CLI or a dummy client (like we did on the bundle tests)

ui/README.md
```diff
@@ -33,3 +33,27 @@ Your app is ready to be deployed!
...
+
+   `./ui/tests/scripts/03-start-grafana.sh`
+
+4. Run the tests with the following command:
+
+    `cd ui && npx playwright test`
+
+You can follow the tests with the ui parameter. This is helpful in case of failures to debug the root cause.
+
+    cd ui && npx playwright test --ui
```
nsklikas, line 59: Let's add a `make e2e` shortcut for this

ui/tests/reset-password.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  const recoveryCode = await getRecoveryCodeFromMailSlurp(context);
+  await page.getByLabel("Recovery code").fill(recoveryCode);
+  await page.getByRole("button", { name: "Submit" }).click();
+
+  await expect(page.getByText("Reset password").first()).toBeVisible();
+  await expect(page).toHaveScreenshot({ fullPage: true, maxDiffPixels: 500 });
+  await page.getByLabel("New password", { exact: true }).fill(USER_PASSWORD_NEW);
+  await page.getByLabel("Confirm New password").fill(USER_PASSWORD_NEW);
+  await page.getByRole("button", { name: "Reset password" }).click();
+
```
nsklikas, line 32: I think we have too much logic here, would it be possible to split it into smaller functions so that we have functions that each handle a single step of the flow and we re-use those functions to test the different permutations of the flow?

ui/tests/use-backup-codes.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  await userPassLogin(page);
+
+  await page
+    .getByRole("button", { name: "Use backup code instead", exact: true })
+    .click();
+
+  await expect(page.getByText("Verify your identity")).toBeVisible();
+  await expect(page).toHaveScreenshot({ fullPage: true, maxDiffPixels: 500 });
+  await page.getByLabel("Backup recovery code").fill(backupCode);
+  await page.getByRole("button", { name: "Sign in", exact: true }).click();
```
nsklikas, line 44: Could this be moved to a separate function?

ui/tests/scripts/01-start-cluster.sh
```diff
@@ -0,0 +1,19 @@
...
+pwd
+
+export KRATOS_PUBLIC_URL="http://localhost:4433"
+export KRATOS_ADMIN_URL="http://localhost:4434"
+export HYDRA_ADMIN_URL="http://localhost:4445"
+export BASE_URL="http://localhost:4455"
+export PORT="4455"
+export TRACING_ENABLED="false"
+export LOG_LEVEL="debug"
+export AUTHORIZATION_ENABLED="false"
```
nsklikas, line 17: This is not needed

ui/tests/scripts/01-start-cluster.sh
```diff
@@ -0,0 +1,19 @@
...
+
+docker-compose -f docker-compose.dev.yml up --build --force-recreate --remove-orphans
```
nsklikas, line 19: We could use `docker compose -f docker-compose.dev.yml  up --remove-orphans --force-recreate`. No need to install the docker-compose binary, the docker CLI has compose built-in.

.github/workflows/e2etest.yaml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          go-version: '1.21'
+      - uses: actions/setup-node@0a44ba7841725637a19e28fa30b79a866c81b0a6 # v4
+        with:
+          node-version: 18
+      - name: Install docker compose
+        run: |
+          sudo curl -L https://github.com/docker/compose/releases/latest/download/docker-compose-$(uname -s)-$(uname -m) -o /usr/local/bin/docker-compose
+          sudo chmod +x /usr/local/bin/docker-compose
+          docker-compose version
```
nsklikas, line 23: Only docker is needed, no need to use the docker-compose binary

.github/workflows/e2etest.yaml
```diff
@@ -0,0 +1,56 @@
...
+          ./ui/tests/scripts/03-start-grafana.sh &
+          sleep 10 # wait for grafana to boot up
+
+      - name: Install Playwright Browser
+        run: cd ui && npx playwright install --with-deps chromium
+
+      - name: Install Kratos binary
+        run: |
+          bash <(curl https://raw.githubusercontent.com/ory/meta/master/install.sh) -d -b . kratos v1.3.0
+          sudo mv ./kratos /usr/local/bin/
```
nsklikas, line 46: We could avoid this by simply using `curl` or running the command from inside the kratos container

ui/tests/helpers/login.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const USER_EMAIL = "test@example.com";
+export const USER_PASSWORD = "test";
+
+export const userPassLogin = async (
+  page: Page,
+  email: string = USER_EMAIL,
+  password: string = USER_PASSWORD,
+) => {
+  await expect(page.getByText("Sign in to grafana")).toBeVisible();
```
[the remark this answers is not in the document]
edlerd: We can use any app, doesn't need to be grafana. But we do need a client application. Do you have a suggestion which app to use as an example client in the e2e tests instead @nsklikas ?

ui/tests/helpers/totp.ts
```diff
@@ -0,0 +1,43 @@
...
+    throw new Error("TOTP setup code not found");
+  }
+  return secretKey;
+};
+
+const setupTotpApp = async (context: BrowserContext, setupCode: string) => {
+  const appName = `app-${randomNameSuffix()}`;
+
+  const totpPage = await context.newPage();
+  await totpPage.goto("https://totp.app/");
```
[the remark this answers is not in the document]
edlerd: Happy to use another solution. Do you have a suggestion what to use to test mfa without using this external service @nsklikas ?

.github/workflows/e2etest.yaml
edlerd: Sure, we can use curl, but how does it work? Can you give me the equivalent of `kratos delete identity  --endpoint http://localhost:4434 $(kratos list identities --endpoint http://localhost:4434 --format json | jq -r .identities[].id)` @nsklikas 
 ?

edlerd: Never mind, this is very straight forward to replace. I figured it out already.

ui/README.md
edlerd: yes, good idea.

ui/tests/helpers/totp.ts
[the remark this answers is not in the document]
nsklikas: No idea right now, I will need to look into it. Let's just open an issue and I will have a look after this is merged

ui/tests/helpers/login.ts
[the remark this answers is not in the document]
nsklikas: There is a hydra CLI command that spins up a client server (see `hydra perform authorization-code`) OR do it like we do in the iam-bundle and not use a client at all. I would suggest to merge it as it is for now, but to open an issue to replace it in the future.

ui/tests/use-backup-codes.spec.ts
edlerd: yes, the whole file can be a bit nicer.

ui/tests/reset-password.spec.ts
edlerd: Yes, agreed. Refactored this to be way shorter and more modular

Makefile
```diff
@@ -49,3 +49,13 @@ npm-build:
 update-openfga-model:
 	@printf 'package authorization\n\n// Code generated by Makefile; DO NOT EDIT.\n\nvar AuthModel = `%s`\n' '$(shell fga model transform --file=./internal/authorization/schema.openfga | jq -c)' > ./internal/authorization/auth_model.go
 .PHONY: update-openfga-model
+
+test-e2e:
+	@echo "Running e2e tests"
+	cd ui && npx playwright test
```
nsklikas, line 55: This should be in the ui makefile and this should call that as in https://github.com/canonical/identity-platform-login-ui/blob/main/Makefile#L45

ui/tests/helpers/totp.ts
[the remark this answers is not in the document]
edlerd: opened an [issue](https://github.com/canonical/identity-platform-login-ui/issues/351)

ui/tests/helpers/login.ts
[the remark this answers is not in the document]
edlerd: opened an [issue](https://github.com/canonical/identity-platform-login-ui/issues/352) for this

ui/tests/scripts/03-start-grafana.sh
edlerd: created an [issue](https://github.com/canonical/identity-platform-login-ui/issues/352) for this.

BarcoMasile (dismissed):

nsklikas (dismissed):
Looks good, thank you

edlerd:
I just found a way to remove the third party service dependency for totp, and rely on the local `oathtool` instead. Added a change for this just minutes ago.

Please another round of review @BarcoMasile and @nsklikas

BarcoMasile (approved):
Great, I'd say let's merge it

nsklikas (approved):
Great! Thanks for that find @edlerd, I think that eventually we should create a container with `oathtool` and not run it on the host namespace. But this can be done later on.

edlerd:
> Great, I'd say let's merge it

[Push The Button](https://www.youtube.com/watch?v=6b9ci_z4v7M) ;)

PR #360, opened 2024-11-13
IAM 1182

shipperizer:
- **ci: only publish via oci-factory in case of a tag release**
- **docs: adding CVE patching instructions**

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,106 @@
+# Security 
+
+
+
+## CVE patching for OCI factory artifacts
+
+When a CVE is reported we are bound to patch the existing OCI artifacts if within the EOL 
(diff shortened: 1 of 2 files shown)


pik4ez-canonical (dismissed):

SECURITY.md
natalian98, line 1: should we merge this doc with https://github.com/canonical/identity-platform-login-ui/blob/main/docs/security.md?

nsklikas (approved):
Don't we need to bump the go version?

wood-push-melon (approved):

SECURITY.md
shipperizer: after today's standup discussion it seems we should keep them separate

PR #372, opened 2024-11-25
chore: upgrade kratos sdk to 1.2

natalian98:
This PR:
- upgrades kratos sdk from 0.13 to 1.2
- upgrades kratos image in the docker setup to 1.3.1
- refactors the unit tests to comply with changes introduced by sdk v1.2
- refactors the UI to keep it consistent (some node labels were renamed).


ui/pages/reset_email.tsx
```diff
@@ -120,6 +121,13 @@ const ResetEmail: NextPage = () => {
     } as RecoveryFlow;
   };
 
+  flow?.ui.nodes.map((node) => {
+    if (isContinueWithPasswordReset(node)) {
+      node.meta.label.text = "Submit";
```
nsklikas, line 126: So we set the text value to `Submit` so that this later gets handled by https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/reset_email.tsx#L100?

Why don't we just change the check there? (IMHO `Continue` is better than `Reset password`, as the user already pressed the reset password button).

ui/components/NodeInputButton.tsx
nsklikas, line 1: After the change in the kratos script, I am not sure if we want to keep the code here as it is. @edlerd should we create a ticket to evaluate this?

nsklikas (commented):
On https://github.com/canonical/identity-platform-login-ui/blob/main/pkg/kratos/service.go#L585 we have a comment about refactoring once we update the sdk version. Is that still valid? Should we open an issue about it or the problem was not solved?

BarcoMasile (approved):

ui/pages/reset_email.tsx
natalian98: I am not sure what https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/reset_email.tsx#L100 does exactly, but it doesn't change the value of the [step 2](https://github.com/canonical/identity-platform-login-ui/blob/main/ui/tests/__screenshots__/reset-password.spec.ts/reset-password-from-oidc-app-2.png) of the recovery flow, which I covered with this small refactor

nsklikas: If I understand correctly it uses the value of the `node.meta.label.text` to make a decision about whether a Back button is needed. AFAICT this is why the tests were originally broken, because the text value was changed from `Continue` to `Submit`.

All I am saying is that I am not sure if we want to add this mapping, just to keep the old texts on the buttons.

ui/components/NodeInputButton.tsx
edlerd: I don't have any context as to what the kratos update changes that would issue a change of this file. What should be changed and why?

nsklikas: [These](https://github.com/ory/kratos/commit/dd6e53d62f343a317edf403218b20599539218c6#diff-465eb3f1ab09546998c26e1e3c42a990488791d1ba402c048285f6f7f71ee348) are the changes to the webauthn.js script.

My question is if we just need to change the names of the functions (like it is done in this PR) or if we should do more changes to our logic. This is not a blocker for this PR, just wondering if we should open an issue to investigate it.

natalian98: Created https://github.com/canonical/identity-platform-login-ui/issues/375

ui/pages/reset_email.tsx
lukasSerelis: The continue button jumping places is not ideal, I'll give a suggestion in tomorrow first thing!

lukasSerelis: This is my proposal:
Back in both pages, resend code as a secondary CTA, as it does not move across pages, and a consistent main CTA to "Reset Password" as that's the main goal of a user going through this flow.
image
image

natalian98: Thanks for the input @lukasSerelis.
I created an issue https://github.com/canonical/identity-platform-login-ui/issues/376 to improve the UI, meanwhile I'll keep the current design to keep the PR focused on the sdk upgrade.

nsklikas (approved):

shipperizer:
> On https://github.com/canonical/identity-platform-login-ui/blob/main/pkg/kratos/service.go#L585 we have a comment about refactoring once we update the sdk version. Is that still valid? Should we open an issue about it or the problem was not solved?

yes please, let's open one

shipperizer (approved):

Issue #371, opened 2024-11-22
Hydra session cookie not created when backup codes are checked


natalian98:
In a scenario when a backup code was used for 2fa and the user has 3 or less codes left, the user is presented with this screen:
![image](https://github.com/user-attachments/assets/1c03d14d-8d2a-4dd6-b4e0-deef26bb309b)
After "I don't need new codes, sign in" is selected, the user is not redirected to grafana as before. Instead, the browser redirects to the login screen again. This happens because the hydra session cookie is not created.

This can be observed when the `shouldRegenerateBackupCodes` check returns true. In other cases (where user has still enough codes), the cookie is created and user is signed in correctly.

We need to replicate the logic introduced for `shouldEnforceMfa` in https://github.com/canonical/identity-platform-login-ui/pull/346/files for the backup codes check to fix this.

PR #387, opened 2024-12-10
fix: do not enforce re-authentication when user was offered to regenerate backup codes

natalian98:
fixes https://github.com/canonical/identity-platform-login-ui/issues/371

### DIFF
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -151,8 +151,9 @@ func (s *Service) MustReAuthenticate(ctx context.Context, hydraLoginChallenge st
 		return true, nil
 	}
 
-	// This is the first user login, they set up their authenticator app, no need to re-auth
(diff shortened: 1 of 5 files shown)


pkg/kratos/service.go
```diff
@@ -156,6 +156,11 @@ func (s *Service) MustReAuthenticate(ctx context.Context, hydraLoginChallenge st
 		return false, nil
 	}
 
+	// Backup code was used for login, no need to re-auth
+	if validateHash(hydraLoginChallenge, c.LoginChallengeHash) && c.BackupCodeUsed {
```
nsklikas, line 160: I think that this `if` can be merged with the one above, no need to calculate the hash 2 times.

nsklikas (commented):
LGTM, only one minor comment.

nsklikas (approved):

PR #392, opened 2024-12-16
fix(login) ensure a password is set, when using password login …

### LINKED ISSUE #382: Incorrect error when trying to sign in without entering a password
If no fields are filled in before "sign in" is pressed, the error shown is irrelevant to the user. 
image
image

### For email should show:
(if incorrect format email address)
image

(if no email entered)
image

### For password should show:
(if empty field)
image

(if password or email is incorrect)
image

edlerd:
to improve error message.

 fixes #382

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -108,13 +108,25 @@ const Login: NextPage = () => {
         }
         return "password";
       };
+      const method = getMethod();
+
+      const isPasswordMissing = !(values as UpdateLoginFlowWithPasswordMethod)
+        .password;
+
+      const setEmptyPassword = () => {
+        (values as UpdateLoginFlowWithPasswordMethod).password = "";
+      };
+
+      if (method === "password" && isPasswordMissing) {
+        setEmptyPassword();
+      }
 
       return kratos
         .updateLoginFlow({
           flow: String(flow?.id),
           updateLoginFlowBody: {
             ...values,
-            method: getMethod(),
+            method,
           } as UpdateLoginFlowBody,
         })
         .then(({ data }) => {


ui/pages/login.tsx
```diff
@@ -103,13 +106,20 @@ const Login: NextPage = () => {
         }
         return "password";
       };
+      const method = getMethod();
+      if (
+        method === "password" &&
+        !(values as UpdateLoginFlowWithPasswordMethod).password
+      ) {
+        (values as UpdateLoginFlowWithPasswordMethod).password = "";
+      }
```
BarcoMasile, line 115: may I suggest we refactor the boolean condition and the action to a function with a "speaking" name? 
So it's clear to anyone what is happening and why

edlerd: added a couple of const to make this easy to read.

ui/pages/login.tsx
```diff
@@ -103,13 +106,25 @@ const Login: NextPage = () => {
...
         return "password";
       };
+      const method = getMethod();
+
+      const isPasswordMissing = !(values as UpdateLoginFlowWithPasswordMethod)
+        .password;
+
+      const setEmptyPassword = () => {
+        (values as UpdateLoginFlowWithPasswordMethod).password = "";
+      };
```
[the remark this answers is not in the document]
edlerd: This is in the submit handler, so it will only run once after user interaction. In JS defining a function cost is near zero. Moving it outside the component will reduce cohesion, so I'd suggest keeping it as it is.

BarcoMasile (dismissed):

edlerd:
rebased on main and resolved conflicts

BarcoMasile (approved):

nsklikas (approved):
LGTM, I tried it and it looks fine. Only question I have is wouldn't it be better to validate the fields before making the request? Not much of a difference tbh.

edlerd:
> LGTM, I tried it and it looks fine. Only question I have is wouldn't it be better to validate the fields before making the request? Not much of a difference tbh.

I think it is obvious that an empty password will not be a valid input that can be submitted. So an error shouldn't be a surprise or cause of frustration as it is expected not to work without a password.

edlerd:
> That's not what I meant.
> 
> In this PR we check if the password is unset and if it is unset we set it to `"'` and let Kratos evaluate the error message. What I meant is that if the password is unset, we just show an error message to the user without making a request to Kratos (thus avoiding the extra request)

I understand your concern was about a possible optimization to avoid sending the request for validation. It is pretty hard to do the validation client side. This is due to the architecture with kratos sending flows of objects. On submit, we can't easily set an error on the password input field, because that is in a different component. We could do it via DOM manipulation. Actually, we do that for in other cases to ease the UX. I am not sure adding this complexity is worth the hassle. Especially as the extra request for validation here should be tiny and quick. I assume in the UX it doesn't make a big difference given a reasonably stable and fast connection.

nsklikas:
Understood, thanks

PR #393, opened 2024-12-16
fix(security-key) improve ux when email does not have a security key set up

### LINKED ISSUE #383: Incorrect error if a security key method is not set up for the email address
When trying to sign in using security key method, after entering email address in the initial sign in screen field, after clicking sign in with security key will simply refresh the page without explaining the error. 

### Flow to reproduce:
Enter an email address into email field (make sure no security keys are set up) -> Click sign in with security key -> You'll be redirected back to starting sign in page with your email field cleared.

### What should happen:
You should see an error saying that this sign in method is unavailable
Enter an email address into email field -> Click sign in with security key -> Email field highlighted and message saying "Sign in method unavailable"
image

edlerd:
fixes #383

### DIFF
--- a/ui/components/NodeInputText.tsx
+++ b/ui/components/NodeInputText.tsx
@@ -25,6 +25,33 @@ export const NodeInputText: FC<NodeInputProps> = ({
   }
   const isDuplicate = deduplicateValues.includes(value as string);
 
+  const message = node.messages.map(({ text }) => text).join(" ");
+  const defaultValue = message.includes("Invalid login method")
+    ? (value as string)
+    : message;
+
+  const getError = () => {
+    if (message.startsWith("Invalid login method")) {
+      return "Invalid login method";
+    }
+
+    if (isDuplicate) {
+      return "This value is already in use";
+    }
+
+    if (
+      attributes.name === "code" ||
+      attributes.name === "totp" ||
+      attributes.name === "totp_code" ||
+      attributes.name === "lookup_secret" ||
+      (isWebauthn && attributes.name === "identifier")
+    ) {
+      return ucFirst(error);
+    }
+
+    return undefined;
+  };
+
   return (
     <Input
       type="text"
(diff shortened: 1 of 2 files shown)


ui/components/NodeInputText.tsx
```diff
@@ -33,17 +42,18 @@ export const NodeInputText: FC<NodeInputProps> = ({
...
+        (attributes.name === "code" ||
         attributes.name === "totp" ||
         attributes.name === "totp_code" ||
         attributes.name === "lookup_secret" ||
         (isWebauthn && attributes.name === "identifier")
           ? ucFirst(error)
           : isDuplicate
             ? "This value is already in use"
-            : undefined
+            : undefined)
```
BarcoMasile, line 56: issue: this is pretty hard to decipher, can we move this behavior in a separate function?

edlerd: Very good idea. I extracted this to make it more readable.

ui/pages/login.tsx
```diff
@@ -179,10 +198,11 @@ const Login: NextPage = () => {
...
     isWebauthn =
-      urlParams.get("webauthn") === "true" ||
-      renderFlow.ui.nodes.filter(
-        (node) => node.group !== "webauthn" && node.group !== "default",
-      ).length === 0;
+      (urlParams.get("webauthn") === "true" ||
+        renderFlow.ui.nodes.filter(
+          (node) => node.group !== "webauthn" && node.group !== "default",
+        ).length === 0) &&
+      !invalid_method;
```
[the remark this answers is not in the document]
edlerd: Yes, also great idea. Making this much easier to gasp with speaking variables.

BarcoMasile (approved):

nsklikas (approved):
Code looks fine, BUT I think that the UX on this feels weird. 

Intuitively the email and password are used when clicking the `Sign in` button. But it turns out that the email is used also when clicking the `Sign in with passkey` button. 

This feels like inconsistent behavior, why don't we also send the email to the idp when the user logs in using oidc?

PR #398, opened 2024-12-20
Support OIDC and WebAuthn sequencing for aal2

natalian98:
This PR adds support for OIDC and WebAuthn sequencing.

### Testing
Follow the [doc](https://github.com/canonical/identity-platform-login-ui/blob/10b9055ed36d67600f4779c6c8b812591843e549/docs/sequencing.md) to set up a testing environment for this case.

image

a. User has no key set up
image

b. User already has a webauthn key
image
The "I want to use another method" link must be removed, I'll open an issue for it.
image

image

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -93,7 +93,7 @@ func serve() {
 		panic("Invalid authorization model provided")
 	}
 
-	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.BaseURL, tracer, monitor, logger)
+	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.OIDCWebAuthnSequencingEnabled, specs.BaseURL, tracer, monitor, logger)
 
 	logger.Infof("Starting server on port %v", specs.Port)
 
--- a/docs/sequencing.md
+++ b/docs/sequencing.md
@@ -1,0 +1,68 @@
+## OIDC-WebAuthn Sequencing
+
+This application exposes an `OIDC_WEBAUTHN_SEQUENCING_ENABLED` environment variable, which defaults to `false`.
+
+**Do not enable this option unless you are sure that this feature applies to your deployment.**
+
+If set to `true`, Login UI will enforce setting up a WebAuthn key (e.g. with YubiKey or Google Password Manager on Android) right after signing in with an external identity provider (Google, GitHub, Entra ID, ...), given that the user has not done it previously.
+Users will be asked to complete it as the second authentication factor on every login flow, regardless of the authentication controls enforced by the identity provider.
+
+Enabling this option works on the assumption that your Kratos config has the `passwordless` flag set to `false`:
+
+```yaml
+selfservice:
+  methods:
+    webauthn:
+      enabled: True
+      config:
+        passwordless: false
+```
+
+This is currently not supported in the Kratos Charmed Operator.
+
+### Testing
+
+In order to test this feature:
+
+1. Switch the aforementioned flag to `false` and disable the `totp` and `password` methods in the [kratos](https://github.com/canonical/identity-platform-login-ui/blob/main/docker/kratos/kratos.yml) configuration for Docker.
+
+2. Set environment variables:
+
+    ```bash
+    export OIDC_WEBAUTHN_SEQUENCING_ENABLED="true"
+    ```
+
+3. Follow the [instructions](https://github.com/canonical/identity-platform-login-ui/blob/main/README.md#try-it-out) to integrate an external identity provider and run the docker setup.
+
+4. Create a hydra client for an application you'll use to test the login flow. Save the client id and secret.
+
+    ```docker
+    docker exec <hydra-container> \
+    hydra create client \
+        --endpoint http://127.0.0.1:4445 \
+        --name <app-name> \
+        --grant-type authorization_code,refresh_token \
+        --response-type code,id_token \
+        --format json \
+        --scope openid,offline_access,email \
+        --redirect-uri <app-redirect-uri>
+    ```
+
+5. Deploy an application that supports OIDC. We'll use Grafana as an example:
+
+    ```docker
+    docker run -d --name=grafana -p 2345:2345 --network identity-platform-login-ui_intranet \
+    -e "GF_SERVER_HTTP_PORT=2345" \
+    -e "GF_AUTH_GENERIC_OAUTH_ENABLED=true" \
+    -e "GF_AUTH_GENERIC_OAUTH_AUTH_ALLOWED_DOMAINS=hydra,localhost" \
+    -e "GF_AUTH_GENERIC_OAUTH_NAME=Identity Platform" \
+    -e "GF_AUTH_GENERIC_OAUTH_CLIENT_ID=<client-id>" \
+    -e "GF_AUTH_GENERIC_OAUTH_CLIENT_SECRET=<client-secret>" \
+    -e "GF_AUTH_GENERIC_OAUTH_SCOPES=openid offline_access email" \
+    -e "GF_AUTH_GENERIC_OAUTH_AUTH_URL=http://localhost:4444/oauth2/auth" \
+    -e "GF_AUTH_GENERIC_OAUTH_TOKEN_URL=http://hydra:4444/oauth2/token" \
+    -e "GF_AUTH_GENERIC_OAUTH_API_URL=http://hydra:4444/userinfo" \
+    grafana/grafana
+    ```
+
+6. Go to the application's login page and click on `Sign in with Identity Platform`. Then, sign in with the integrated identity provider. Upon a successful authentication, you will be asked to create a WebAuthn key.
--- a/pkg/extra/handlers.go
+++ b/pkg/extra/handlers.go
@@ -11,7 +11,6 @@ import (
 	client "github.com/ory/kratos-client-go"
 
 	"github.com/canonical/identity-platform-login-ui/internal/logging"
-	httpHelpers "github.com/canonical/identity-platform-login-ui/internal/misc/http"
 	"github.com/canonical/identity-platform-login-ui/pkg/kratos"
 )
 
@@ -19,10 +18,10 @@ type API struct {
 	service ServiceInterface
 	kratos  kratos.ServiceInterface
 
-	logger                        logging.LoggerInterface
 	baseURL                       string
 	oidcWebAuthnSequencingEnabled bool
 	contextPath                   string
+	logger                        logging.LoggerInterface
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
@@ -42,6 +41,12 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 	}
 
 	consentChallenge := r.URL.Query().Get("consent_challenge")
+	if consentChallenge == "" {
+		err = fmt.Errorf("no consent challenge present")
+		a.logger.Errorf(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
 
 	if a.oidcWebAuthnSequencingEnabled {
 		// enforce webauthn setup
@@ -55,9 +60,28 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 
 		if shouldEnforceWebAuthn {
 			returnTo, _ := url.JoinPath("/", a.contextPath, "/ui/consent")
-			returnToConsent, _ := httpHelpers.AddParamsToURL(returnTo, httpHelpers.QueryParam{Name: "consent_challenge", Value: consentChallenge})
-			a.webAuthnSettingsRedirect(w, returnToConsent)
+			returnToConsent, err := url.ParseRequestURI(returnTo)
+			if err != nil {
+				return
+			}
+
+			q := returnToConsent.Query()
+			q.Set("consent_challenge", consentChallenge)
+			returnToConsent.RawQuery = q.Encode()
+			err = a.webAuthnSettingsRedirect(w, returnToConsent.String())
+			if err != nil {
+				err = fmt.Errorf("unable to build webauthn redirect path, possible misconfiguration, err: %v", err)
+				a.logger.Error(err.Error())
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+			}
+
 			return
+		}
+
+		if session.GetAuthenticatorAssuranceLevel() == "aal1" {
+			err = fmt.Errorf("webauthn step was skipped, user has not completed 2fa")
+			a.logger.Error(err.Error())
+			http.Error(w, err.Error(), http.StatusForbidden)
 		}
 	}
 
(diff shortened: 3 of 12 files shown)


docs/sequencing.md
```diff
@@ -0,0 +1,69 @@
+## OIDC-WebAuthn Sequencing
+
+This application exposes an `OIDC_WEBAUTHN_SEQUENCING_ENABLED` environment variable, which defaults to `false`.
+
+**Do not enable this option unless you are sure that this feature applies to your deployment.**
+
+If set to `true`, Login UI will enforce setting up a WebAuthn key (e.g. with YubiKey or Google Password Manager on Android) right after signing in with an external identity provider (Google, GitHub, Entra ID, ...), given that the user has not done it previously.
+Users will be asked to complete it as the second authentication factor on every login flow, regardless of the authentication controls enforced by the identity provider.
+
+Enabling this option assumes that your Kratos config sets the `passwordless` flag to `false`:
```
BarcoMasile, line 10:
```suggestion
Enabling this option  works on the assumption that your Kratos config has the `passwordless` flag set to `false`:
```

pkg/extra/handlers.go
```diff
@@ -63,13 +92,67 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 
 }
 
-func NewAPI(service ServiceInterface, kratos kratos.ServiceInterface, logger logging.LoggerInterface) *API {
+func (a *API) shouldEnforceWebAuthnWithSession(ctx context.Context, session *client.Session) (bool, error) {
+	hasOIDC := false
+	// enforce only if one of the authentication methods was oidc
+	for _, method := range session.AuthenticationMethods {
+		if method.Method != nil && *method.Method == "oidc" {
+			hasOIDC = true
```
wood-push-melon, line 100: I guess we can early break the loop once we confirm `hasOIDC` is `true`.

wood-push-melon: (OPTIONAL) And also we could move the following logic inside the `if` block. If OIDC method is confirmed, we can directly check `a.kratos.HasWebAuthnAvailable`.

pkg/extra/handlers.go
```diff
@@ -63,13 +92,67 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
+
+	return !webAuthnAvailable, nil
+}
+
+func (a *API) webAuthnSettingsRedirect(w http.ResponseWriter, returnTo string) {
+	redirect, err := url.JoinPath("/", a.contextPath, "/ui/setup_passkey")
+	if err != nil {
+		err = fmt.Errorf("unable to build webauthn redirect path, possible misconfiguration, err: %v", err)
+		a.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
wood-push-melon, line 121: nitpick:

(I might be wrong and pedantic, feel free to ignore). Personally I think it's better to return HTTP error response in the handler method instead of inside a private method.

pkg/kratos/service.go
```diff
@@ -798,6 +798,48 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
...
+		return false, err
+	}
+
+	var (
+		webauthnInfo kClient.IdentityCredentials
+		ok           = false
+	)
+
+	if webauthnInfo, ok = identity.GetCredentials()["webauthn"]; !ok {
+		return false, nil
```
wood-push-melon, line 817: question: is it better to add a log here? (for future debugging) because we have multiple places returning `false, nil`

pkg/kratos/service.go
```diff
@@ -798,6 +798,48 @@ func (s *Service) HasTOTPAvailable(ctx context.Context, id string) (bool, error)
...
+	}
+
+	credentialsSlice, ok := webauthnInfo.Config["credentials"].([]interface{})
+	if !ok {
+		// user has no webauthn keys
+		return false, nil
+	}
+
+	for _, credentialElem := range credentialsSlice {
```
wood-push-melon, line 826: I guess we could do something like the following to save one nested block:

```
for _, cred := range credentials {
    credential, ok := cred.(map[string]interface{})
    if !ok {
        continue
    }
    
    isPasswordless, ok := credential["is_passwordless"]
    if ok && !isPasswordless.(bool) {
        s.logger.Debugf("Identity %s has a 2FA WebAuthn key", id)
        return true, nil
    }
}
```

cmd/serve.go
```diff
@@ -93,7 +93,7 @@ func serve() {
 		panic("Invalid authorization model provided")
 	}
 
-	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.BaseURL, tracer, monitor, logger)
+	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.OIDCWebAuthnSequencingEnabled, specs.BaseURL, tracer, monitor, logger)
```
wood-push-melon, line 96: (Definitely NOT in this PR's scope), but I guess we should start to note that this argument list keeps expanding. I think we'd better refactor this in the future with some idiomatic way.

BarcoMasile: Thank you for pointing this out, I agree 100%
My old proposal for this was to use a builder-pattern like approach to make it cleaner, instead of relying on a composite structure. But I am ok with anything that is better than this

pkg/extra/handlers.go
```diff
@@ -32,8 +41,28 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			err = fmt.Errorf("webauthn enforce check error: %v", err)
+			a.logger.Error(err.Error())
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		if shouldEnforceWebAuthn {
+			returnTo, _ := url.JoinPath("/", a.contextPath, "/ui/consent")
+			returnToConsent, _ := httpHelpers.AddParamsToURL(returnTo, httpHelpers.QueryParam{Name: "consent_challenge", Value: consentChallenge})
```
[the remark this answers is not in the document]
nsklikas: I would suggest not to over-complicate it. If you feel that this is ugly, perhaps a better alternative would be to change the function to:
```
AddParamToURL(url, name, value string) (string, error)
```

pkg/extra/handlers.go
```diff
@@ -63,13 +92,67 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
+	// Set the original consent URL as return_to, to continue the flow after WebAuthn key is set
+	redirectTo, _ := httpHelpers.AddParamsToURL(redirect, httpHelpers.QueryParam{Name: "return_to", Value: returnTo})
+
+	w.WriteHeader(http.StatusSeeOther)
+	_ = json.NewEncoder(w).Encode(
+		kratos.ErrorBrowserLocationChangeRequired{
+			Error:             &client.GenericError{Id: &errorId},
+			RedirectBrowserTo: &redirectTo,
+		},
+	)
```
nsklikas, line 136: IMO we should either respond with a 30x and set the `Location` in the header OR respond with a 40x (maybe 422 as kratos does?) and set the redirect location in the response body. Considering this is a backchannel request I would go with the latter.

nsklikas (commented):
Other than the race condition I don't see any major problems with the code, but I have a more general comment/issue.

What is implemented here is not what is described in the spec. The spec says that this functionality will be similar with the existing totp enforcing functionality. BUT the implementation is different than the previous logic, as the enforcing happens on the consent step rather than on the login step. The race condition mentioned on my comment is caused by this. Other than that, this implementation does not make much sense from a higher level. We accept the login request and then on the consent step, we make the actual decision for whether the login is valid or not.

I have no problem with validating in the consent step that the session has the required aal level (this should be configurable). But imo we should enforce registration on the login step.

I suppose that you chose to go with this approach because kratos redirects straight to the consent endpoint so you don't have a chance to intercept the request like we do for password logins. The reason for the different behavior is that for password logins we update the kratos flow using a `POST` request, which results in kratos telling us to redirect in the response body, but in the OIDC flow the user returns kratos with a `GET` and kratos can redirect the user to hydra on its own by over-riding the return_to (see https://github.com/ory/kratos/blob/master/selfservice/flow/registration/hook.go#L309, https://github.com/ory/kratos/blob/master/selfservice/flow/login/hook.go#L270)

IMHO we should stop sending the login_challenge to kratos and handle the hydra request entrirely on the login UI.

Finally, I don't like that you have to set 2 flags to make this work. Setting `MFA_ENABLED=False` does not make much sense as we are enforcing MFA on OIDC.

pkg/extra/handlers.go
BarcoMasile: Keep in mind this is not a backchannel request since it comes from a frontend useragent and it's returned to it
This is also aligned with how we've done redirect in both admin ui and login ui

pkg/extra/handlers.go
```diff
@@ -32,8 +41,28 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	consentChallenge := r.URL.Query().Get("consent_challenge")
+
+	if a.oidcWebAuthnSequencingEnabled {
```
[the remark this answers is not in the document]
BarcoMasile: Great catch!!
I agree this could happen, but since it could happen only once and it would need to be caught at the exact split second in between these two invocations, I would really consider it an acceptable risk since it's extremely unlikely to happen, my opinion would be to disregard this risk as not likely to happen. I'm curious to see what the rest of the team thinks

pkg/extra/handlers.go
```diff
@@ -32,8 +41,28 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
+		// enforce webauthn setup
+		shouldEnforceWebAuthn, err := a.shouldEnforceWebAuthnWithSession(r.Context(), session)
```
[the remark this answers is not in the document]
natalian98: The problem with this approach is that in case a user signs in with a backup code as 2fa for any reason (method != webauthn), but already has a passkey tied to the account, we would have to decide between forcing to set up another security key or assuming that the user has one. When fetching the identity credentials, we make sure that a key is set regardless of the method used for 2fa.

I could partly adopt your suggestion and exit quicker in case the aal2 method is webauthn, but I'm not sure that'd be a clean approach
```go
if a.oidcWebAuthnSequencingEnabled && (*session.AuthenticatorAssuranceLevel == "aal1" || *session.AuthenticationMethods[1].Method != "webauthn") {
        // enforce webauthn setup
        shouldEnforceWebAuthn, err := a.shouldEnforceWebAuthnWithSession(r.Context(), session)
```

pkg/extra/handlers.go
nsklikas: Ok maybe I abused the term back-channel. But it is an XHR request, meaning that it does not make sense to return a redirect. Our approach right now is not consistent, in some places we return `200`, in others `303` and in others `403`.

Out of the 3 I think that `303` is the one that feels most un-intuitive. But we can move it to a new issue to implement a uniform strategy for that.

pkg/extra/handlers.go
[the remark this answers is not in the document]
nsklikas: That's why I suggest to enforce in the login part of the flow. That way in the consent we would only care about having aal2.

pkg/extra/handlers.go
BarcoMasile: Definitely
IMO 303 would still be the most "appropriate", but I see what you mean. Probably 200 or 422, but let's move it to an issue. We'll have to address the smaller things at some point

pkg/extra/handlers.go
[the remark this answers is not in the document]
nsklikas: I agree that it is an edge case, but it may lead to other vulnerabilities and since there is a workaround, I see no reason not to implement it.

cmd/serve.go
shipperizer: - please feel free to open an enhancement issue and we can tackle it next pulse

pkg/extra/handlers.go
```diff
@@ -32,8 +41,34 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	consentChallenge := r.URL.Query().Get("consent_challenge")
```
BarcoMasile, line 44: question: are we good even if the value is empty here? or should we add the check and behave differently?

pkg/extra/handlers.go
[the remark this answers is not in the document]
natalian98: Switched to use standard lib as suggested in https://github.com/canonical/identity-platform-login-ui/pull/398#discussion_r1908639476

cmd/serve.go
natalian98: https://github.com/canonical/identity-platform-login-ui/issues/414

pkg/extra/handlers_test.go
```diff
@@ -46,7 +48,68 @@ func TestHandleConsentSuccess(t *testing.T) {
 	mockService.EXPECT().AcceptConsent(gomock.Any(), *session.Identity, consent).Return(accept, nil)
 
 	mux := chi.NewMux()
-	NewAPI(mockService, mockKratosService, mockLogger).RegisterEndpoints(mux)
+	NewAPI(mockService, mockKratosService, mockLogger, BASE_URL, false).RegisterEndpoints(mux)
+
+	mux.ServeHTTP(w, req)
```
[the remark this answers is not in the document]
natalian98: https://github.com/canonical/identity-platform-login-ui/issues/415

natalian98:
> What is implemented here is not what is described in the spec. The spec says that this functionality will be similar with the existing totp enforcing functionality. BUT the implementation is different than the previous logic, as the enforcing happens on the consent step rather than on the login step.

The user experience is similar, but I had to change the implementation compared to totp because the oidc flow differs from the "kratos-only" ones. The session is set after the oidc callback and as you pointed out the last kratos call before the consent is a `GET`, so I could not find a way to intercept it while already having a session.

> IMHO we should stop sending the login_challenge to kratos and handle the hydra request entrirely on the login UI.

The [frontend logic](https://github.com/canonical/identity-platform-login-ui/blob/b4ff004b2cf561485047ecb6921ed0f9ceadfd6d/ui/pages/login.tsx#L73) relies on `login_challenge` presence for kratos flows, I'd say that refactoring it is out of scope of this PR but as discussed yesterday let's create a spike for it.

> Finally, I don't like that you have to set 2 flags to make this work. Setting MFA_ENABLED=False does not make much sense as we are enforcing MFA on OIDC.

Actually this flag doesn't have to be disabled. There is a check on the totp enforcement preventing from making users set it up if the sign in method was oidc, so the 2 features don't conflict. I updated the readme removing this requirement.

Also the `MFA_ENABLED` naming that doesn't fully reflect its actual purpose. Setting this flag to true enforces to set up a totp method for 2fa specifically. I suppose we agreed on that name due to the UX design for standard cases, without considering that we may want to make the users set up another 2fa method. Perhaps it should be renamed to `ENFORCE_TOTP_MFA` or similar.
@nsklikas

pkg/extra/handlers.go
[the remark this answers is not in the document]
nsklikas: I think that after the `a.oidcWebAuthnSequencingEnabled` if block, we should add this check:
```
...
if a.oidcWebAuthnSequencingEnabled {
...
}

if (oidcWebAuthnSequencingEnabled OR mfaEnabled)  && session.AuthenticatorAssuranceLevel == "aal1"{
    return 403 # or something like this
}
...
```

This will also work as a word around for the race condition.

pkg/extra/handlers.go
```diff
@@ -63,13 +110,66 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
 
+	a.baseURL = baseURL
+	a.oidcWebAuthnSequencingEnabled = oidcWebAuthnSequencingEnabled
+
+	fullBaseURL, err := url.Parse(baseURL)
+	if err != nil {
+		// this should never happen if app is configured properly
+		a.logger.Fatalf("Failed to construct API base URL: %v\n", err)
+	}
+	a.contextPath = fullBaseURL.Path
```
nsklikas, line 172: We pass the `baseURL` and convert it to `contextPath`. We store both variables in the struct, but only contextPath is used. I think we should keep only one of the 2.

Since baseURL is used in the kratos module as well, I would suggest to keep only that to stay consistent

internal/misc/http/helpers.go
```diff
@@ -45,3 +51,17 @@ func GetUserClaims(i kratos_client.Identity, cr hydra_client.OAuth2ConsentReques
 
 	return ret
 }
+
+func AddParamsToURL(u string, qs ...QueryParam) (string, error) {
```
nsklikas, line 55: This was moved, but it still only used in the kratos module, not sure if we want to revert this change or not.

Also I think we should either use this function everywhere or throw it away (see https://github.com/canonical/identity-platform-login-ui/pull/398/files#diff-eee000f25821f185296fc9109030f323fca69b610a1b5ef787071dca62a8cf38R141, @shipperizer do you want to get rid of this function?)

nsklikas (commented):
lgtm, just a couple nit-picks

pkg/extra/handlers.go
natalian98: Both `baseURL` and `contextPath` are stored in the struct in [kratos](https://github.com/canonical/identity-platform-login-ui/blob/8acccc89e48df4946921f518a182c546220a8a74/pkg/kratos/handlers.go#L28-L29C2) module, so I think it's already consistent

pkg/extra/handlers.go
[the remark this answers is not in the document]
natalian98: We don't want to return a 403 error in aal1 case. That just means the user has completed oidc, but has to create a key to get the session elevated to aal2.
In the other case where a key is present, kratos prevents from the race condition by requiring aal2 to get the session, as the highest available factor has to be completed first.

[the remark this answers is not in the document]
nsklikas: The user won't be able to get there if they are redirected to register webauthn.

The only way for someone to get in that if clause is if:
1. session has aal=aal1
2. identity has 2nd fa registered (because otherwise `shouldEnforceWebAuthn=true` and the user would have been redirected to register their key)

This means that the user has registered webauthn, but hasn't completed 2fa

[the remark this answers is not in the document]
natalian98: To sum up what we discussed after the daily:
User won't get a session until 2fa is completed IF a key is registered, that is restricted by [`highest_available`](https://github.com/canonical/identity-platform-login-ui/blob/8acccc89e48df4946921f518a182c546220a8a74/docker/kratos/kratos.yml#L23-L25) requirement so kratos prevents from this. But I will add the check to validate if the aal requirement is respected. This will catch any edge case and ensure that aal2 is always required, even if for any reason kratos breaks the current logic or the setting is misconfigured.

wood-push-melon (approved):

BarcoMasile (approved):

nsklikas (approved):
LGTM, I didn't test it again but before the last few changes everything was good. Thank you for the good work and managing so many reviews at the same time.

PR #400, opened 2025-01-07
ci: build on ubuntu 22.04 and fix e2e tests

natalian98:
This PR fixes the CI failures:
- `ubuntu-latest` was recently changed to 24.04, that is causing issues with rockcraft build: https://github.com/canonical/identity-platform-login-ui/actions/runs/12648123440/job/35241836957
We can use 22.04 instead.
- updates the e2e tests screenshots to reflect on the recent change: `E-mail` -> `Email`


natalian98:
I needed to update 2 of the e2e tests screenshots to make the CI pass through. There was a small change that causes them to fail: `E-mail` --> `Email`
Let me know if you think those commits should be splitted into 2 PRs, we will need to bypass branch protection rules though.

BarcoMasile:
> I needed to update 2 of the e2e tests screenshots to make the CI pass through. There was a small change that causes them to fail: `E-mail` --> `Email` Let me know if you think those commits should be splitted into 2 PRs, we will need to bypass branch protection rules though.

@natalian98 
Let's keep these here, but change the title and the description of the PR so we don't lose the info

edlerd (approved):

PR #412, opened 2025-01-09
Ensure white background on cards

edlerd:
# Done

- ensure white background on cards
- update screenshots


BarcoMasile (approved):
Approved. I have to revive the same comment I made when these screenshot based tests were first added: I thought they would bring pain, it seems that they are.

Should we reconsider this?

edlerd:
> Approved. I have to revive the same comment I made when these screenshot based tests were first added: I thought they would bring pain, it seems that they are.

I strongly disagree, they are useful. They stopped us from releasing with a broken layout / wrong background on the cards. See [discussion in mattermost](https://chat.canonical.com/canonical/pl/w6aycdaaejywf8nbyiefoyk51y).

BarcoMasile:
> I strongly disagree, they are useful. They stopped us from releasing with a broken layout / wrong background on the cards. See [discussion in mattermost](https://chat.canonical.com/canonical/pl/w6aycdaaejywf8nbyiefoyk51y).

I was referring to also an issue we had a few days ago in this week, same problem.
Let's reassess in a few weeks then.

PR #436, opened 2025-01-23
feat(ui) improve sequenced login ux

edlerd:
# Done
- improve UX for sequenced login
- auto redirect to OIDC provider if it is the only login option
- add checkbox to automatically open the passkeys dialogue on landing on next login. Storing the state in local storage.

To-do: blocked by receiving the status from the backend if sequenced login is enabled

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -4,7 +4,7 @@ import {
   UiNodeInputAttributes,
   UpdateLoginFlowBody,
 } from "@ory/client";
-import { Spinner } from "@canonical/react-components";
+import { CheckboxInput, Spinner } from "@canonical/react-components";
 import { AxiosError } from "axios";
 import type { NextPage } from "next";
 import { useRouter } from "next/router";
@@ -20,11 +20,32 @@ import {
   UpdateLoginFlowWithOidcMethod,
   UpdateLoginFlowWithPasswordMethod,
 } from "@ory/client/api";
-import { isSignInEmailInput, isSignInWithPassword } from "../util/constants";
+import {
+  isSignInEmailInput,
+  isSignInWithHardwareKey,
+  isSignInWithPassword,
+} from "../util/constants";
+import {
+  isWebauthnAutologin,
+  toggleWebauthnSkip,
+} from "../util/webauthnAutoLogin";
 
 const Login: NextPage = () => {
   const [flow, setFlow] = useState<LoginFlow>();
+  const [isSequencedLogin, setSequencedLogin] = useState(false);
   const isAuthCode = flow?.ui.nodes.find((node) => node.group === "totp");
+
+  useEffect(() => {
+    void fetch("/api/v0/app-config")
+      .then((response) => {
+        const data = response.json();
+        return data as { oidc_webauthn_sequencing_enabled?: boolean };
+      })
+      .then((data) => {
+        setSequencedLogin(data.oidc_webauthn_sequencing_enabled ?? false);
+      })
+      .catch(console.error);
+  }, []);
 
   // Get ?flow=... from the URL
   const router = useRouter();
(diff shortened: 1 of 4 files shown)


ui/pages/login.tsx
```diff
@@ -299,13 +308,77 @@ const Login: NextPage = () => {
...
   });
 
+  // automatically forward to single oidc provider if it is the only option
+  const isSingleOidcOption =
+    isSequencedLogin &&
+    renderFlow?.ui.nodes.length === 2 &&
+    renderFlow?.ui.nodes[1].group === "oidc" &&
+    (renderFlow?.ui.nodes[0].attributes as UiNodeInputAttributes).name ===
+      "csrf_token";
+  if (isSingleOidcOption) {
```
BarcoMasile, line 324: How about a function with this logic so the if can become:

```typescript
  // outside of the component
  function isSingleOidcOption(isSequencedLogin boolean, renderFlow renderFlowType) {
      return isSequencedLogin &&
          renderFlow?.ui.nodes.length === 2 &&
          renderFlow?.ui.nodes[1].group === "oidc" &&
          (renderFlow?.ui.nodes[0].attributes as UiNodeInputAttributes).name === "csrf_token";
  }
  
  ...
  
  if (isSingleOidcOption(isSequencedLogin, renderFlow)) {
      ...
   }
```

ui/pages/login.tsx
```diff
@@ -299,13 +308,77 @@ const Login: NextPage = () => {
...
+    );
+  };
+
+  const toggleWebauthnSkip = () => {
+    if (isWebauthnAutologin()) {
+      localStorage.removeItem(WEBAUTHN_AUTOLOGIN_KEY);
+    } else {
+      localStorage.setItem(WEBAUTHN_AUTOLOGIN_KEY, WEBAUTHN_AUTOLOGIN_VALUE);
+    }
+  };
```
BarcoMasile, line 348: I do appreaciate the clarity of these statements!
Since these are all based on constants, these functions should be outside of the component

ui/util/constants.ts
```diff
@@ -44,5 +48,10 @@ export const isSignInEmailInput = (node: UiNode): node is NodeWithLabel =>
...
   node.meta.label?.id === ORY_LABEL_SIGN_IN_WITH_PASSWORD;
 
-export const isContinueWithPasswordReset = (node: UiNode): node is NodeWithLabel =>
+export const isContinueWithPasswordReset = (
+  node: UiNode,
+): node is NodeWithLabel =>
   node.meta.label?.id === ORY_LABEL_CONTINUE_PASSWORD_RESET;
+
+export const isSignInWithHardwareKey = (node: UiNode): node is NodeWithLabel =>
+  node.meta.label?.id === ORY_LABEL_SIGN_IN_WITH_HARDWARE_KEY;
```
BarcoMasile, line 57: nitpick(feel free to disregard): these two and the `isSignInWithPassword` look like they would belong to a different file

BarcoMasile (commented):
I know it's a draft, I wanted to point out a couple of things

nsklikas (commented):
question: why use `localStorage` instead of `SessionStorage`? If I remember correctly localstorage is persisted across sessions and sessionstorage is local only to the tab. Are we expecting these values to never change? (imagine a case where we fetch more config values from the backend, or where the backend is misconfigured and a variable changes, or just that we decide to change a value after an upgrade)

ui/pages/login.tsx
edlerd: I usually try to avoid callbacks when the logic can be directly evaluated. It seems simpler and a bit cheaper to me.

edlerd:
> question: why use `localStorage` instead of `SessionStorage`? If I remember correctly localstorage is persisted across sessions and sessionstorage is local only to the tab. Are we expecting these values to never change? (imagine a case where we fetch more config values from the backend, or where the backend is misconfigured and a variable changes, or just that we decide to change a value after an upgrade)

It is a local setting for the user that should be in the browser, not in the server. This is similar to the "remember me" setting on login, it is client specific.

ui/util/constants.ts
edlerd: I think they all belong here. Only the webauthn constants wouldn't fit, I moved them into another place just now.

BarcoMasile (approved):

nsklikas:
> It is a local setting for the user that should be in the browser, not in the server. This is similar to the "remember me" setting on login, it is client specific.

I was referring to the browser session storage, but I misunderstood what you were storing there. Please disregard.

PR #447, opened 2025-01-27
feat: add endpoint to expose application info

natalian98:
- adds a `GET /api/v0/app-config` endpoint exposing the `OIDC_WEBAUTHN_SEQUENCING_ENABLED` flag value

### DIFF
--- a/pkg/status/handlers.go
+++ b/pkg/status/handlers.go
@@ -22,8 +22,13 @@ type Health struct {
 	Hydra  bool `json:"hydra"`
 }
 
+type DeploymentInfo struct {
+	OidcSequencingEnabled bool `json:"oidc_webauthn_sequencing_enabled"`
+}
+
 type API struct {
-	service ServiceInterface
+	oidcWebAuthnSequencingEnabled bool
+	service                       ServiceInterface
 
 	tracer tracing.TracingInterface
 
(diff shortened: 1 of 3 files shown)


pkg/status/handlers.go
```diff
@@ -35,6 +40,7 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/status", a.alive)
 	mux.Get("/api/v0/version", a.version)
 	mux.Get("/api/v0/ready", a.ready)
+	mux.Get("/api/v0/info", a.info)
```
BarcoMasile, line 43: suggestion: `"/api/v0/app-config"`

pkg/status/handlers_test.go
```diff
@@ -122,3 +122,34 @@ func TestHealthFailure(t *testing.T) {
...
+	mockTracer := NewMockTracingInterface(ctrl)
+	mockService := NewMockServiceInterface(ctrl)
+
+	req := httptest.NewRequest(http.MethodGet, "/api/v0/info", nil)
+	w := httptest.NewRecorder()
+
+	mux := chi.NewMux()
+	NewAPI(false, mockService, mockTracer, mockMonitor, mockLogger).RegisterEndpoints(mux)
+
+	mux.ServeHTTP(w, req)
```
BarcoMasile, line 141: can we switch to the "essential" unit test, relying on the specific endpoint instead of instantiating a Mux ?

BarcoMasile:
Also, I think the PR title should be: add endpoint to expose application info

nsklikas (approved):
lgtm, my only question is: is this the correct handler to put this? I would expect it to be in the UI handler (I suppose it will be used only by the UI). I'm fine with leaving it as it is

BarcoMasile (approved):

PR #463, opened 2025-02-03
Setup stale action

BarcoMasile:
## Description
Setup the [Stale action](https://github.com/actions/stale) to run every work day at 7 AM.
The current configuration allows for 60 day before labeling an issue as "stale", and 10 days before closing it.

Right now it will consider only issues, and only ones created after today.
This is to avoid disrupting our status quo.

If this is accepted then I'll proceed with the Admin UI repo as well.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,23 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '0 7 * * 1-5' # every morning at 7 AM, only work days
+
+permissions:
+  issues: write
+
+jobs:
+  stale:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: 60
+          stale-issue-message: "This issue has been marked as stale due to the lack of updates in the last 60 days"
+          days-before-close: 7
+          close-issue-message: "This issue has been closed due to the lack of updates"
+          stale-issue-label: 'stale'
+          exempt-issue-labels: 'high-priority'
+          stale-pr-label: ''  # Don't mark PRs as stale
+          exempt-pr-labels: '*'  # Exclude PRs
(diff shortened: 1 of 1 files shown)


shipperizer (approved):

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/stale@v9
+        with:
+          days-before-stale: 60
+          stale-issue-message: "This issue has been marked as stale due to the lack of updates in the last 60 days"
+          days-before-close: 7
+          close-issue-message: "This issue has been closed due to the lack of updates"
```
natalian98, line 19:
```suggestion
          close-issue-message: "This issue has been closed due to the lack of updates"
          close-issue-reason: 'not_planned'
```

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,24 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '0 7 * * 1-5' # every morning at 7 AM, only work days
+
+permissions:
+  contents: write
```
natalian98, line 7: I think this is needed for delete-branch option, but we don't want the stale bot to close PRs

wood-push-melon (approved):

.github/workflows/stale.yaml
BarcoMasile: This will be the message that will appear as a comment in the issue, why would you prefer it that way?

.github/workflows/stale.yaml
BarcoMasile: You're right, I'll remove it

.github/workflows/stale.yaml
natalian98: Just to know that the issue has not been worked on.
I am fine with keeping just the message though.

natalian98 (approved):

PR #465, opened 2025-02-04
Refactor webauthn 2fa

nsklikas:
IAM-1295

As described in IAM-1279, this PR stops sending the `login_challenge` to kratos if sequencing is enabled. It also introduces the following changes:
- Checks in the consent page that the session has the required aal, this acts as a fail-safe for any misconfiguration in Kratos.
- Refactors how webauthn happens as a 2nd FA, previously the login UI backend was not involved when submitting the assertion, unlike all other 2fa methods. Not sure if this is the case also for 1st FA (passkeys) and if any changes are needed for that as well (cc @natalian98).
- Adds a new config to `app-config`

@edlerd please have a look at the changes to `login.tsx`, feel free to push any changes to my branch or I can just create an issue for you to refactor it afterwards. Can we change it so that the form submission triggers a back-channel call, so that we can handle it as we do for other flow0s? Also we should change it so that the data is sent as a JSON instead of FORM (IMO easiest approach would be to use the kratos sdk)


natalian98:
Overall the webauthn as 2fa functionality works well, except that "Another layer of authentication before you get access to" is now missing the app name - is it not sent to frontend?
missing_app_name

Some bugs were introduced in other flows:

1. When signing in with webauthn as 1fa (passwordless=True), user doesn't get logged in (hydra session is not created), instead is redirected to a page with `redirect_to`:
```
{"redirect_to":"http://localhost:4433/self-service/login/browser?aal=aal2\u0026return_to=http%3A%2F%2Flocalhost%3A4455%2Fui%2Flogin%3Flogin_challenge%3DIewEHi9UcdX-PCjDAXUXWp5Y0W2VqBc6UZD8cs5BQYkXM07IyBNeBW0loy_wsU_ycgVhfidR5Cjx0Zp5OD6W3CZtHkE18bFvnFnDq7JjHmx3PYaYnq-xCno2Q2v-Get3Ms6kxTk6dcqBjlVrkp7qHe-5GMr_lXBmbNxS-5RV8vwppcmWTa9tnH6t1-YflUN-UXAjAbJElxk23RxQrgbrDtFRGdtE1h41cPdfi0vgliKjd6jtBb_eRJJFbPZgS6jR3hErdPxuTH1oBYKH1LkQTQzXU0D9GEJmerJ53w09x21d_f3CAxN-UqK0qIMhL2vrxwiktZBa_50H_k7riYyZ8DLbrBCGpUixEBjxB9TvJr9DTVBX1HUK-BXN1VnJmOmcwoLuDtTGQI33q5t_5W4NRenWDyiQuEbaNAnV6Qj6DWdpWvAgxhiOlx8CasZ6wpOctZrcVzLpCCa66oA3id8pplJDqRrban4-DvOH1nBxLoOY7B3oxVkdEELUnFqFvoWxa6Ngw3qfQr_i9c2E6VVPzGwglhxklEsN1_ornAA4YJE8AZTbbzSZiDoWbCBNu8Sz7-BZd2QXJupAd4GNy3XBFvi0YjcrzoE17lJm0WGdv9gZG8sPng48W9JN9LpODn9wwaASxhtvZUa8g39auTYzWXuu6a0iD3bmOJIgznJZQin9sVvJLHEIVpSjCI5fOjeZc5ZIsJqsnq9LgAElT2LqiyzbUMIG_NKl6pR7iy0uR09Sd2zJL1ldh6S2tayPFdhkxGF3vizQC7JX4sX3ajRgNwmW9zwlhO5IKMenbY35_8OAR_NP5PjY--1-lkQfnuY0nRWFBhr_vzMJgbLsXw94tUCgDMwsXe8k-bpchuaOvKNVy3pIabr4qdjA7ezkGqq0JXGU3EV8oCq0IPxbMq9DgeuNf1zh26r7rcaXbGeAlYDYKafPVM5f7xbgTRgnnoBwPdTXcnTJpvOyqLBsOSAguAYUPxNusNiWhP0_nbUc4x2XJNNTYo-TRV7fQD35vW1PtEiVU4a-5N2px1kpHYon4fXhP1rNNkrn-fLLT8_7-dwTiulTH5BkOYpghgzlGSK0YFQ9zCgCFekfqKJffVtueE3iZfRYpC3sKI793g9BOLn4Am20MmP4lors1k0zWuVaTDJLIKF27MDIoSoxtFmG5jbNQydtMJB0DikDLqJIF3dzLV3HKrDUrdtD07c6KoKnJRxTw990WIRLazaChBAAYNfqHWJ328S-fNow9BrlYKc3ApXNuYNN_d2Y1BNYU-LH-FNtyT3szSOM67zHy-yAvXqyFm8ivBxc6Nj39ohhuV5FojsDP6jZMTTnbR7uXzhc2BjcLVuZ2O--rI6dpXvNzpRKyUljpBdqDG_6DtpFcYC9DOGmJvbI-rsnU-ZEDxa5jrL0-uJKWKA0jcobBq2PUZNEtuBAba68YB04EYKHUm4dW7Gu5qNcfbb84tL7gbTcMEByHBhBvrjtOZ9YzV5K5Ux8lxVKqB8G61Y2nUxd70PEp5rT-XcsSpm3vljDEm-K5_6deojwh5hdKUUdjO6Hxiq_G8uhiPT5AQ8tdA%253D%253D"}
```

2. When resetting a password, on authentication code step the sign in fails with `"failed to accept login request: 400 Bad Request"`.
reset_password_failure

ui/pages/login.tsx
```diff
@@ -297,6 +302,12 @@ const Login: NextPage = () => {
       });
 
       return null;
+    } else if (isWebauthn) {
+      // TODO: Refactor me
+      const u = renderFlow.ui.action.split("self-service");
+      if (u.length == 2) {
+        renderFlow.ui.action = baseURL + "/api/kratos/self-service" + u[1];
+      }
```
edlerd, line 310: This probably shouldn't be in the else branch?
```suggestion
    }
    if (isWebauthn) {
      // TODO: Refactor me
      const u = renderFlow.ui.action.split("self-service");
      if (u.length == 2) {
        renderFlow.ui.action = baseURL + "/api/kratos/self-service" + u[1];
      }
```

edlerd: You probably want to [split on first occurrence](https://stackoverflow.com/questions/4607745/split-string-only-on-first-instance-of-specified-character) and not split and hope that you only get one match to drop the `u.length == 2` condition.

edlerd: Also probably best to match "/self-service/", to avoid matching parts of the domain.

ui/pages/login.tsx
```diff
@@ -33,16 +33,21 @@ import {
 const Login: NextPage = () => {
   const [flow, setFlow] = useState<LoginFlow>();
   const [isSequencedLogin, setSequencedLogin] = useState(false);
+  const [baseURL, setBaseURL] = useState("");
```
edlerd, line 36: I don't think we need this. Can't we use `window.location.origin` below instead?

nsklikas: this will omit the path of the URL, in juju deployments the baseURL will be `https://some/path/` and in local deployments it will be `http://localhost:4455`. IMO This is an easy workaround that allows us to be flexible

ui/pages/login.tsx
edlerd: Not sure how the SDK would help here.

ui/pages/login.tsx
edlerd: I think we know which path we are on, because this file is login under the pages directory. If we strip the /login?... from the end of the current url, that should be a safe way to calculate this, while preserving paths in the base url.

ui/pages/login.tsx
nsklikas: The sdk is already configured with the correct URL and will send the payload in JSON instead of FORM. This custom logic wouldnt be required if we used the sdk to make the request.

edlerd: Ah yes, this would also save us from setting up a base url state (see the other comment). Just not sure if we can get the correct url out of it. Because here we need to set the form target url and not issue a request directly.

ui/pages/login.tsx
nsklikas: That's true, I still believe this is the most flexible approach though as it does not make any assumptions on the relative position of the paths, thus I prefer it. Do you see any problems with this?

Also I just realized that (most likely) the request to the `app-config` endpoint will not work on juju deployments, as the path `/api/v0/app-config` is wrong, it should leverage the base_url instead (cc @natalian98).

ui/pages/login.tsx
nsklikas: Could we add a listener for the `submit` event to this form so that we can override the default behavior and use the sdk instead?

nsklikas: Otherwise I guess we will just have to handle it in the backend.

edlerd: > Could we add a listener for the submit event to this form so that we can override the default behavior and use the sdk instead?

That is possible, but seems a bit hacky.

I think it will be better to construct the base url in the same way we do it for the kratos sdk. We could make that logic reusable. Then constructing the prefixed form target url similar you do it here already.

nsklikas:
@natalian98 thanks for testing it, all of your issues should be resolved now

pkg/kratos/service.go
```diff
@@ -897,21 +933,52 @@ func (s *Service) HasNotEnoughLookupSecretsLeft(ctx context.Context, id string)
...
 		return true
+	case "webauthn":
+		return !s.oidcWebAuthnSequencingEnabled
 	default:
 		return false
 	}
 }
 
-func NewService(kratos KratosClientInterface, kratosAdmin KratosAdminClientInterface, hydra HydraClientInterface, authzClient AuthorizerInterface, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {
+func (s *Service) hydrateKratosLoginFlow(ctx context.Context, flow *kClient.LoginFlow) error {
```
BarcoMasile, line 945: suggestion: this method should only do the hydration of the login flow. The `if` check should be done outside of this method, in the calling function so that
```go
  if s.oidcWebAuthnSequencingEnabled {
    s.hydrateKratosLoginFlow( ... )
  }

```
Also, it would make the method simpler and more predictable in "what it does"

BarcoMasile (commented):
Left a few comments and a couple of questions for my clarity.
Great work overall

pkg/extra/handlers.go
```diff
@@ -116,50 +85,17 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
...
-	// Set the original consent URL as return_to, to continue the flow after WebAuthn key is set
-	redirectTo, err := url.ParseRequestURI(redirect)
-	if err != nil {
-		return err
+// sessionRequiredAAL returns the required aal, based on the session's authentication methods.
+func (a *API) sessionRequiredAAL(session *kClient.Session) kClient.AuthenticatorAssuranceLevel {
+	if a.oidcWebAuthnSequencingEnabled && len(session.AuthenticationMethods) >= 1 && session.AuthenticationMethods[0].GetMethod() == "oidc" {
+		return kClient.AUTHENTICATORASSURANCELEVEL_AAL2
+	} else if a.mfaEnabled && len(session.AuthenticationMethods) >= 1 && session.AuthenticationMethods[0].GetMethod() == "password" {
+		return kClient.AUTHENTICATORASSURANCELEVEL_AAL2
```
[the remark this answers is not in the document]
natalian98: `session.AuthenticationMethods[0].GetMethod()` can also be `webauthn` apart from oidc and password

ui/pages/login.tsx
natalian98: I see no reason for it to be unreachable on juju, but I tried it to make sure and the endpoint is accessible:
```
$ curl http://10.64.140.43/test2-identity-platform-login-ui-operator/api/v0/app-config
{"oidc_webauthn_sequencing_enabled":false}
```

natalian98: @edlerd does [this](https://github.com/canonical/identity-platform-login-ui/blob/798330ab0210e4f1ed5ef763cfa20e4da2511fef/ui/pages/login.tsx#L40) call respect the base url?
On juju deployments it's similar to:
```
$ curl http://10.64.140.43/test2-identity-platform-login-ui-operator/api/v0/app-config
{"oidc_webauthn_sequencing_enabled":false}
```

pkg/kratos/handlers.go
```diff
@@ -147,12 +165,13 @@ func (a *API) handleCreateFlowNewSession(r *http.Request, aal string, returnTo s
...
 }
 
-func (a *API) handleCreateFlowWithSession(r *http.Request, session *client.Session, loginChallenge string) (any, []*http.Cookie, error) {
+func (a *API) handleCreateFlowWithSession(w http.ResponseWriter, r *http.Request, session *client.Session, loginChallenge string) (*hClient.OAuth2RedirectTo, []*http.Cookie, error) {
 	response, cookies, err := a.service.AcceptLoginRequest(r.Context(), session, loginChallenge)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to accept login request: %v", err)
 	}
 
+	a.cookieManager.ClearStateCookie(w)
```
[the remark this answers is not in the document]
nsklikas: Good point, will move it to caller

pkg/kratos/handlers.go
```diff
@@ -258,20 +281,62 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+		_ = json.NewEncoder(w).Encode(response)
+	}
+
+	resp, err := flow.MarshalJSON()
+	if err != nil {
+		a.logger.Errorf("Error when marshalling Json: %v\n", err)
+		http.Error(w, "Failed to parse flow error", http.StatusInternalServerError)
+		return
+	}
+	setCookies(w, cookies)
```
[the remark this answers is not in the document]
nsklikas: You are right, this was on accident. Good catch.

pkg/extra/handlers.go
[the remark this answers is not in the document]
nsklikas: Will try to refactor it to a `switch` statement. Let me know if it is clearer, otherwise I can switch to using functions as you suggested

natalian98:
The issues from yesterday were resolved, thanks.

Small request:
Could you remove the redundant "to": https://github.com/canonical/identity-platform-login-ui/blob/798330ab0210e4f1ed5ef763cfa20e4da2511fef/ui/pages/login.tsx#L370 as it's already in `getTitleSuffix`?
toto

This was not changed in your PR, but since you are already refactoring the login page we could include this small fix.

pkg/kratos/handlers.go
```diff
@@ -258,20 +281,62 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
 
-	setCookies(w, cookies)
-
 	if shouldRegenerateBackupCodes {
-		a.lookupSecretsSettingsRedirect(w, flowId, *loginFlow.ReturnTo, loginFlow.GetOauth2LoginChallenge())
+		a.lookupSecretsSettingsRedirect(w, flowId, *loginFlow.ReturnTo, flowCookie)
 		return
 	}
 
+	setCookies(w, cookies)
```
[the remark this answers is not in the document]
natalian98: If we are redirecting user to regenerate backup codes it means they already used one and have a sufficient aal, so I think it's fine

pkg/extra/handlers.go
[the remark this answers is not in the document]
nsklikas: > `session.AuthenticationMethods[0].GetMethod()` can also be `webauthn` apart from oidc and password

If it is `webauthn`, we have no flag for enforcing aal2 so no specific logic is needed for that.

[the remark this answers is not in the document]
nsklikas: > question(involving also @natalian98 for this): are we 100% sure we can safely rely on the order in which methods populate the `session.AuthenticationMethods` slice?

We are already making that assumption on https://github.com/canonical/identity-platform-login-ui/blob/main/pkg/kratos/handlers.go#L300. I changed it to loop over all the AMRs, not sure if I should revert it or not.

natalian98:
I found another issue, similar to the one from yesterday.
1. sign in with pwd, set up totp
2. set up a passkey
3. unlink totp
4. in a new browser session, sign in with passkey
5. no hydra session created, user redirected to 
```
{"error":{"id":"session_aal2_required","message":""},"redirect_browser_to":"/ui/setup_secure?return_to=http%3A%2F%2Flocalhost%3A4455%2Fui%2Flogin%3Flogin_challenge%3DYXJXsLPBr4rh2KUBRGbiyMFLVEiGOqIuSu8wrvD5PiuigsBItTmV9-IsmFR3xcC7bgJF_-azrKQ4-YgwRg9v5m6qvn140cDhANHSStZBok9jit2O1Q_baW6ws2FXjKAryckzelvXN8DJ-3bTPiKKdAwY1YFd1Axxg9B49qIF8BmYtDG9-bV27iMK-7bimxiVb0b_jv3Og1qrlPef7NBLQCJTzbHb2N5SFqEfSiBFdUcIDx5mGFrYDMQy0uq5ZOYXxMkAS5rG2yL9nNjvNBtDrSlknKlLkXauuj5rCdn6woa7BoZgVX31O3NUgwMW73omsVsZJz2IusX_o2xuCmw1puyJ50CFG69j83ebz9jYqWitln2DRRfiUVBWPx-rRDh2pGWCCF9B4Mb1r6SWQXpODO55sieukHkGm2mCcM1adnkMeu2ksCDla4zxZn1veuzqZJFYKcXnra1HGOeNtlgTDkQ5F7ZTjYsnoJZGyEsFg8CtncezEtvXfvpjNPrxZVtSWI1Kx0u7KGQEekYjtlOuqVqIiwBFDSaAB-pjiC4pfnuhGimmk09NH9RtYH8nCYOev3ph8plUbBHKr7Mtzz25-XZkpkXPrY_CfrCDnzRz1KY6HImQPPzfyQdXFaG117naMEaRCpbWIB1DzP1VlY-i9GlzrYcmjs6gtxSUU94uhXfypbBFt-PmDyjUOHq0iAnrNqK-Z4qCZHcmSwAIf0cEvGegMhV2IdS-jDlCTW0mJ6H8ErIa-ohcMl4p5JJxw8HEH4_9JnoEgR2PfAA-XPfB3tO5c5ifsmqKwDc8ThLwnz7NbgfVeH0qZYsFoPr7ylA2MW2L2MIR9aaW5ord9liDrpdjim9-tqMthboM9d-bNel-Lof4iKl3I3nXOApAXZxQGXlH1j_glLR5oCM81llagSFc_CUOoel1YNOdalY1pZIf1k-SXraxz5dsW6m_mjc7365isbHTIrRrkcVzzUF0z3C2tSI6a-snhiFY9XpZvVsqkbRdRNI3toyLVk2gRiG-lx83d9zmuGHY92v3FnscCsd8-yiRO9Smqv3Xy7682lxz-Il7jAQ8XNCdzlRXMafHFxQkNIHyVILq28mxwxvx7-Jp0z8gvKRRbS4zQh-LjNFxOdGuBoLztdd_SX9YFOfy4wgXwbKVhjCpcizFIWsV17vnCHifXkKbmmuE9Ipd4ew2HRn0VDbfW46iv5JTf1dlN7mTP1k3q1BCOYqXcdN8sQdYozk2nX9aqAIXbzvO9SDlumDPz08isWonvNhUh5iEFHl-uiL40tTGco8db5PZj9qROLv8_dMfP8_HGzP3ub06TRdJiQU91qoZLD1vYEatd9DaRMkxjvaf8enXfwGR4nnP8Igk0HutFYa7IknN5XOvMIggYE3IDxag1sI37kMkWr11127tAPmTjqAbvhtwCPS2giyIDrtC5RvqxqzRmy0_o-wbSUoG1AH9J8TBsckyyukyKuExoa40cdkshvcXLLTyP10mHpuoEj759llYc8QMI_hvjGRTzgXrAPQ45fduyUS0q1litLeb0OJWFYhotrNn2tq5vJY-Q0x5-_4KUFRLwg%253D%253D"}
```

Passwordless `webauthn` is aal1 so 2fa should still be enforced

pkg/extra/handlers.go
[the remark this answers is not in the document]
natalian98: Passwordless `webauthn` is aal1 so kratos still enforces 2fa in that case. TOTP setup is also enforced (see https://github.com/canonical/identity-platform-login-ui/blob/ce4a352bce73bd62f74512cb4f7f0eebce5df51a/pkg/kratos/handlers.go#L340, it is only skipped for oidc)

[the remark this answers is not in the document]
nsklikas: You are right, thank you!

nsklikas:
> I found another issue, similar to the one from yesterday.
> 
>     1. sign in with pwd, set up totp
> 
>     2. set up a passkey
> 
>     3. unlink totp
> 
>     4. in a new browser session, sign in with passkey
> 
>     5. no hydra session created, user redirected to

Good catch. All these issues seem to be caused by the change to handle webauthn in the login UI backend, which are unrelated to the original scope of the PR.

My latest commit should handle this, as I wrote in the TODO, I will refactor this logic for IAM-1286.

The PR seems to be growing out of control, if more issues arise I think I will break it into multiple smaller PRs.

BarcoMasile (dismissed):
Great work

natalian98 (approved):
All of the aforementioned issues were solved, thank you.

@edlerd could you please address the remaining comments that concern the frontend part?

pkg/extra/handlers.go
```diff
@@ -40,6 +39,12 @@ func (a *API) handleConsent(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if session.GetAuthenticatorAssuranceLevel() < a.sessionRequiredAAL(session) {
+		a.logger.Errorf("insufficient session aal, this indicates a misconfiguration on kratos")
+		http.Error(w, "insufficient session aal", http.StatusForbidden)
```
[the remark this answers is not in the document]
nsklikas: Fair point, I always forget about their difference. IMO 403 is correct here, as there is a user session but it does not have enough aal.

Maybe 401 is better for the error above, where we fail to fetch a valid user session

pkg/kratos/handlers.go
```diff
@@ -234,13 +255,17 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+	redirectTo, flow, cookies, err := a.service.UpdateLoginFlow(r.Context(), flowId, *body, cookies)
 	if err != nil {
 		a.logger.Errorf("Error when updating login flow: %v\n", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
+	u, _ := url.Parse(loginFlow.GetReturnTo())
+	lc := u.Query().Get("login_challenge")
+	flowCookie := FlowStateCookie{LoginChallengeHash: hash(lc)}
```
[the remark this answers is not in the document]
nsklikas: We don't care. The hash will always be the same, but we won't use the cookie for non hydra logins (`lc == "" => not a hydra login`). We could avoid computing the hash if there is no login_challenge, but I didn't think it was worth to add an extra branch to the code.

ui/pages/login.tsx
edlerd: > @edlerd does [this](https://github.com/canonical/identity-platform-login-ui/blob/798330ab0210e4f1ed5ef763cfa20e4da2511fef/ui/pages/login.tsx#L40) call respect the base url? On juju deployments it's similar to:
> 
> ```
> $ curl http://10.64.140.43/test2-identity-platform-login-ui-operator/api/v0/app-config
> {"oidc_webauthn_sequencing_enabled":false}
> ```

I think that call will not work well on juju deployments.

pkg/kratos/handlers.go
```diff
@@ -261,17 +286,66 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
 
 	w.WriteHeader(http.StatusOK)
-	_ = json.NewEncoder(w).Encode(flow)
+	w.Write(resp)
+}
+
+// TODO: Refactor me for IAM-1286
+func (a *API) backchannelRedirectResponse(w http.ResponseWriter, r *http.Request, resp RedirectToInterface) {
```
shipperizer, line 332: random q: why `backchannel`?

pkg/kratos/service.go
```diff
@@ -178,17 +206,24 @@ func (s *Service) CreateBrowserLoginFlow(
...
 
 	flow, resp, err := request.Execute()
 	if err != nil {
 		return nil, nil, err
 	}
 
+	// Populate the flow with the hydra login req, so that the UI can retrieve this info
+	if s.oidcWebAuthnSequencingEnabled {
+		s.hydrateKratosLoginFlow(ctx, flow)
+	}
```
shipperizer, line 225: having this here after the clause has been evlauted in https://github.com/canonical/identity-platform-login-ui/pull/465/files#diff-5e4ba31969008b305d67765d5d4c9bd50d9bcc49571f1594e97d2b152ddf1378R209 seems a but strange

is there anything we can reuse from the request instead of this?would that be cleaner ?

edlerd (approved):
LGTM

The frontend needs a bit of changes, but those will be done in a follow up:
The base url is a bit of a chick/egg problem. Because we are calling the api endpoint to get the url to the api. So i think we should construct it in the ui regardless. Probably best to abstract this and reuse in all places. I'll try to come up with something and will likely drop the "fetch base url" approach later on.

pkg/kratos/service.go
```diff
@@ -897,21 +955,50 @@ func (s *Service) HasNotEnoughLookupSecretsLeft(ctx context.Context, id string)
...
 		return true
+	case "webauthn":
+		return !s.oidcWebAuthnSequencingEnabled
 	default:
 		return false
 	}
 }
 
-func NewService(kratos KratosClientInterface, kratosAdmin KratosAdminClientInterface, hydra HydraClientInterface, authzClient AuthorizerInterface, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {
+func (s *Service) hydrateKratosLoginFlow(ctx context.Context, flow *kClient.LoginFlow) error {
```
shipperizer, line 967: this method is changing the flow via reference, wouldn't it be better to return a new flow with the enhanced/hydrated attributes?

pkg/kratos/handlers.go
nsklikas: Wasn't sure what name to use. I wanted this to be used to handle redirect responses for XHR requests (which are back-channel requests), BUT given the limitations of our UI this will have to handle browser redirects as well. I will rename it.

pkg/kratos/service.go
nsklikas: I tried to refactor it, should be better now.

BarcoMasile (approved):

shipperizer (approved):

Issue #399, opened 2024-12-20
Handle errors by error id, not status code


natalian98:
The consent page doesn't use the [flow error handlers](https://github.com/canonical/identity-platform-login-ui/blob/main/ui/util/handleFlowError.ts), unlike other pages ([see login](https://github.com/canonical/identity-platform-login-ui/blob/main/ui/pages/login.tsx), for example).
We need to switch from checking the status code to error id.

PR #490, opened 2025-02-17
Uniform POST response redirection

nsklikas:
IAM-1286

Use `ErrorBrowserLocationChangeRequired` only in the service layer, this is a always converted to `BrowserLocationChangeRequired` before being returned to the handler. Only `BrowserLocationChangeRequired` should be used in the handler from now on. The frontend now has to handle only `BrowserLocationChangeRequired` responses.

All redirect responses to XHR will now return `200 OK`, also thought about using `422` to have similar behavior to kratos. IMO It does not really matter as long as we have a consistent behavior in backend-frontend communications. (cc @natalian98 if you have a suggestion/preference in this)

### DIFF
--- a/pkg/kratos/handlers_test.go
+++ b/pkg/kratos/handlers_test.go
@@ -293,7 +293,7 @@ func TestHandleCreateFlowRedirectToSetupWebauthn(t *testing.T) {
 		t.Errorf("expected error to be nil got %v", err)
 	}
 
-	if res.StatusCode != http.StatusSeeOther {
+	if res.StatusCode != http.StatusOK {
 		t.Fatalf("expected HTTP status code 303 got %v", res.StatusCode)
 	}
 	loginFlow := ErrorBrowserLocationChangeRequired{}
--- a/pkg/kratos/handlers_test.go
+++ b/pkg/kratos/handlers_test.go
@@ -294,14 +294,14 @@ func TestHandleCreateFlowRedirectToSetupWebauthn(t *testing.T) {
 	}
 
 	if res.StatusCode != http.StatusOK {
-		t.Fatalf("expected HTTP status code 303 got %v", res.StatusCode)
+		t.Fatalf("expected HTTP status code 200 got %v", res.StatusCode)
 	}
-	loginFlow := ErrorBrowserLocationChangeRequired{}
+	loginFlow := BrowserLocationChangeRequired{}
 	if err := json.Unmarshal(data, &loginFlow); err != nil {
 		t.Errorf("expected error to be nil got %v", err)
 	}
-	if !strings.HasPrefix(*loginFlow.RedirectBrowserTo, "/ui/setup_passkey") {
-		t.Errorf("expected redirect_browser_to to start with '/ui/setup_passkey' got %v", *loginFlow.RedirectBrowserTo)
+	if !strings.HasPrefix(*loginFlow.RedirectTo, "/ui/setup_passkey") {
+		t.Errorf("expected redirect_to to start with '/ui/setup_passkey' got %v", *loginFlow.RedirectTo)
 	}
 }
 
@@ -315,7 +315,7 @@ func TestHandleCreateFlowWithSession(t *testing.T) {
 
 	session := kClient.NewSession("test")
 	redirect := "https://some/path/to/somewhere"
-	redirectTo := hClient.NewOAuth2RedirectTo(redirect)
+	redirectTo := BrowserLocationChangeRequired{RedirectTo: &redirect}
 
 	loginChallenge := "login_challenge_2341235123231"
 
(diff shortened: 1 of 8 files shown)


BarcoMasile (dismissed):
LGTM, although I disagree with choosing 200 OK as status code for redirection since 302 is specifically for redirecting responses, but if the team is ok with that specific change I will not slow down the PR so:
approved

nsklikas:
No problem with any status code, only thing that matters is that we have a consistent pattern across FE and BE. Reason I don't like 302/303 is that they are used to redirect a request to a different location (see https://www.rfc-editor.org/rfc/rfc9110#status.302 and https://www.rfc-editor.org/rfc/rfc9110#status.303) and that's the reason we were not setting the `Location` header before (because that would break the request).

IMO the response we send to the FE is: "your request was handled successfully and now you need to redirect the user-agent to this URL", that's why I chose to go with `200`. If it wasn't an XHR request then I would have probably used 303 (with the `Location` header).

pkg/kratos/handlers_test.go
```diff
@@ -293,15 +293,15 @@ func TestHandleCreateFlowRedirectToSetupWebauthn(t *testing.T) {
 		t.Errorf("expected error to be nil got %v", err)
 	}
 
-	if res.StatusCode != http.StatusSeeOther {
+	if res.StatusCode != http.StatusOK {
 		t.Fatalf("expected HTTP status code 303 got %v", res.StatusCode)
```
natalian98, line 297:
```suggestion
		t.Fatalf("expected HTTP status code 200 got %v", res.StatusCode)
```

same for https://github.com/canonical/identity-platform-login-ui/blob/af13afdceaf73850c086e0d0f26762e62fa41694/pkg/kratos/handlers_test.go#L658

natalian98 (dismissed):
Ok it makes sense, thanks for giving more explanation. I just have a minor comment on unit tests.

Other than that frontend needs to be refactored (see https://github.com/canonical/identity-platform-login-ui/blob/5f1df12b0a4001d2caa7696c1d090253de2d3ac3/ui/pages/consent.tsx#L48 for example, we should handle the errors by id instead of status code, as mentioned in [this](https://github.com/canonical/identity-platform-login-ui/issues/399) issue.

pkg/kratos/handlers_test.go
nsklikas: fixed, thank you

nsklikas:
> Other than that frontend needs to be refactored (see
> 
> https://github.com/canonical/identity-platform-login-ui/blob/5f1df12b0a4001d2caa7696c1d090253de2d3ac3/ui/pages/consent.tsx#L48
> for example, we should handle the errors by id instead of status code, as mentioned in [this](https://github.com/canonical/identity-platform-login-ui/issues/399) issue.

I didn't want to make too many changes, in order not to break anything. I can try to go through the frontend and change those as well if you think that it should be part of this PR.

About the linked code of `consent.tsx`. I think that was part of the initial sequencing implementation. I don't think that can happen anymore (because 2fa registration is enforced in the login handler). Do you see any problem with removing it altogether? I will try to see if it breaks something tomorrow.

natalian98:
I am fine with refactoring frontend separately, as long as the feature doesn't get broken.

You are right, it [does not depend](https://github.com/canonical/identity-platform-login-ui/blob/603acacdc12c29a0646972f5e73d04f9003884f0/pkg/kratos/handlers.go#L336) on the 303 status code anymore, so it can probably be removed. I'll try running some tests as well to confirm that.

pkg/kratos/service.go
```diff
@@ -1011,6 +967,22 @@ func (s *Service) hydrateKratosLoginFlow(ctx context.Context, flow *kClient.Logi
...
+		return nil, err
+	}
+
+	// We trasform the kratos response to our own custom response here.
+	// The original kratos response contains an 'Error' field, which we remove
+	// because this is not a real error.
+	return &BrowserLocationChangeRequired{
+		RedirectTo: redirectResp.RedirectBrowserTo,
+		Error:      redirectResp.Error,
+	}, nil
```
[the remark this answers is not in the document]
nsklikas: Good catch. I copied/pasted the comment, but it was probably wrong/outdated. I removed the comment now.

BarcoMasile (approved):

nsklikas:
> I found one issue that needs a refactor in [error handlers](https://github.com/canonical/identity-platform-login-ui/blob/main/ui/util/handleFlowError.ts).
> 
> When you reset password for an account with 2fa, after inserting a recovery code, user is not redirected to complete the second factor.
> 
> After receiving 403 from http://localhost:4455/api/kratos/self-service/settings/flows?id=..., browser redirects to an unexisting page: http://localhost:4455/ui/undefined
> 
> In the logs: `"Failed to get settings flow due to insufficient aal: &{0xc0001d0370 0xc000264dc0}\n"}`
> 
> I guess this is because `BrowserLocationChangeRequired` returns `redirect_to` instead of `redirect_browser_to`.

I had missed one of the backend responses and it was still returning an error. I pushed some more changes, now the backend returns 403 if Kratos responded with that, otherwise it's a 200. Reason is that there were some paths in the frontend that relied on the response being an error and I didn't want to have to do a FE refactor. I tried several flows and couldn't find any issues now.

BarcoMasile (approved):

natalian98 (approved):

PR #492, opened 2025-02-18
Fix webauthn for local deployments

### LINKED ISSUE #355: Webauthn script not loaded on localhost due to Content Security Policy violation
The passwordless login method is broken in the latest login ui version.
To reproduce:
1. Run the docker compose with login ui image >v0.17.0, create a hydra client and run grafana.
2. Go to localhost:2345, log in with the test user.
3. Go to localhost:4455/ui/setup_passkey. Insert any key name and click on Add. The console will display the following errors:
```
Refused to load the script 'http://localhost:4433/.well-known/ory/webauthn.js' because it violates the following Content Security Policy directive: "script-src 'self' 'unsafe-inline' 'unsafe-eval'". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback.

framework-2c16ac744b6cdea6.js:9 Uncaught TypeError: r.__oryWebAuthnRegistration is not a function
    at onClick (127-3320e06beedeceb7.js:1:2103)
    at Object.eU (framework-2c16ac744b6cdea6.js:9:14908)
    at eH (framework-2c16ac744b6cdea6.js:9:15062)
    at framework-2c16ac744b6cdea6.js:9:33368
    at re (framework-2c16ac744b6cdea6.js:9:33467)
    at rn (framework-2c16ac744b6cdea6.js:9:33881)
    at framework-2c16ac744b6cdea6.js:9:39337
    at oN (framework-2c16ac744b6cdea6.js:9:96304)
    at eF (framework-2c16ac744b6cdea6.js:9:14046)
    at ro (framework-2c16ac744b6cdea6.js:9:35172)
```

Due to this error, the webauthn.js script is not loaded and the QR code pop-up is not shown.

This seems to be caused by https://github.com/canonical/identity-platform-login-ui/pull/299 (I tried removing the changes it introduced and the script was loaded correctly).

nsklikas:
IAM-1187

Closes #355 

Introduces a new flag that allows localhost:* for the script-src.

I removed `unsafe-inline` from the `script-src`, I could not remove it from the `style-src` because it is required from some library. (cc @edlerd). IMHO we should remove `unsafe-inline` and explicitly specify the hashes of the scripts that we allow to use inline, I would expect this to be handled at the proxy rather than the app itself.

### DIFF
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/fs"
 	"net/http"
+	"net/url"
 	"path"
 	"strings"
 
@@ -17,7 +18,9 @@ const UI = "/ui"
 type API struct {
 	fileServer http.Handler
 
-	dev bool
+	baseURL         string
+	csp             string
+	kratosPublicURL string
 
 	logger logging.LoggerInterface
 }
@@ -38,7 +41,7 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 		w.Header().Set("X-Frame-Options", "SAMEORIGIN")
 		// Sets the Content Security Policy (CSP) for the page, which helps mitigate XSS attacks and data injection attacks.
 		// The policy allows loading resources (scripts, styles, images, etc.) only from the same origin ('self'), data URLs, and all subdomains of ubuntu.com.
-		w.Header().Set("Content-Security-Policy", a.getCSP())
+		w.Header().Set("Content-Security-Policy", a.csp)
 
 		// `no-store`: This will tell any cache system not to cache the index.html file
 		// `no-cache`: This will tell any cache system to check if there is a newer version in the server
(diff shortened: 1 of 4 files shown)


nsklikas:
Decided not to use a new flag and just compare the 2 URLs, this seems cleaner.

pkg/ui/handlers.go
```diff
@@ -62,11 +68,24 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
 	a.fileServer.ServeHTTP(w, r)
 }
 
-func NewAPI(fileSystem fs.FS, logger logging.LoggerInterface) *API {
+func (a *API) getCSP(baseURL, kratosPublicURL string) string {
+	b, _ := url.Parse(baseURL)
+	k, _ := url.Parse(kratosPublicURL)
+	if k != nil && b != nil && k.Host != b.Host {
```
nsklikas, line 74: Parsing the URLs may fail, this is swallowed here because the `BaseURL` is not a required parameter. Since this happens when the application starts, we could also panic to force the user to provide valid config.

pkg/ui/handlers.go
```diff
@@ -62,11 +68,24 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
...
+		// We need to whitelist the kratos URL, because the browser needs to fetch the webauthn.js script from Kratos
+		return fmt.Sprintf("default-src 'self' data: https://*.ubuntu.com; script-src 'self' %v; style-src 'self' 'unsafe-inline'", kratosPublicURL)
+	}
+	return "default-src 'self' data: https://*.ubuntu.com; script-src 'self'; style-src 'self' 'unsafe-inline'"
```
BarcoMasile, line 78: suggestion: feel free to disregard the code suggestion but consider the edit to the comment to comply with company guidelines

```suggestion
        additionalScriptURL := ""
	if k != nil && b != nil && k.Host != b.Host {
		// Allowlist the kratos URL to allow the browser needs to fetch the webauthn.js script
		additionalScriptURL := kratosPublicURL
	}
	return fmt.Sprintf("default-src 'self' data: https://*.ubuntu.com; script-src 'self' %v; style-src 'self' 'unsafe-inline'", additionalScriptURL)
```

nsklikas: Oh nice, that's better. Thank you

natalian98 (approved):

BarcoMasile (approved):

edlerd (approved):
Good move forward to make the CSP a bit tighter.

PR #494, opened 2025-02-19
Use relative path for api access in login.tsx to simplify

edlerd:
# Done
* ensure api call to identify sequencing is enabled works when deployed within a subfolder
* simplify base url for kratos api override

fixes WD-19160

### DIFF
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -30,24 +30,22 @@ import {
   toggleWebauthnSkip,
 } from "../util/webauthnAutoLogin";
 
+type AppConfig = {
+  oidc_webauthn_sequencing_enabled?: boolean;
+};
+
 const Login: NextPage = () => {
   const [flow, setFlow] = useState<LoginFlow>();
   const [isSequencedLogin, setSequencedLogin] = useState(false);
-  const [baseURL, setBaseURL] = useState("");
   const isAuthCode = flow?.ui.nodes.find((node) => node.group === "totp");
 
   useEffect(() => {
-    void fetch("/api/v0/app-config")
+    void fetch("../api/v0/app-config")
       .then((response) => {
-        const data = response.json();
-        return data as {
-          oidc_webauthn_sequencing_enabled?: boolean;
-          base_url?: string;
-        };
+        return response.json() as Promise<AppConfig>;
       })
       .then((data) => {
         setSequencedLogin(data.oidc_webauthn_sequencing_enabled ?? false);
-        setBaseURL(data.base_url ?? "");
       })
       .catch(console.error);
   }, []);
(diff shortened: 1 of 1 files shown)


ui/pages/login.tsx
```diff
@@ -33,21 +33,16 @@ import {
...
   useEffect(() => {
-    void fetch("/api/v0/app-config")
+    void fetch("../api/v0/app-config")
       .then((response) => {
         const data = response.json();
-        return data as {
-          oidc_webauthn_sequencing_enabled?: boolean;
-          base_url?: string;
-        };
+        return data as { oidc_webauthn_sequencing_enabled?: boolean };
```
nsklikas, line 42: suggestion: would it make sense to define an object with the `app-config` fields?

edlerd: added a type for this.

nsklikas (approved):

ui/pages/login.tsx
```diff
@@ -30,24 +30,22 @@ import {
...
+
 const Login: NextPage = () => {
   const [flow, setFlow] = useState<LoginFlow>();
   const [isSequencedLogin, setSequencedLogin] = useState(false);
-  const [baseURL, setBaseURL] = useState("");
   const isAuthCode = flow?.ui.nodes.find((node) => node.group === "totp");
 
   useEffect(() => {
-    void fetch("/api/v0/app-config")
+    void fetch("../api/v0/app-config")
```
BarcoMasile, line 43: question: are you sure this works with any context path the UI could be deployed at?

mas-who (commented):
Just one small query, also how do we go about testing this?

ui/pages/login.tsx
edlerd: This is a relative path. We use the same approach in other places already. It should always work [tm], because this component is a page and we can make the assumption that '../' relates to the base url.

ui/pages/login.tsx
```diff
@@ -308,10 +306,10 @@ const Login: NextPage = () => {
 
       return null;
     } else if (isWebauthn) {
-      // TODO: Refactor me
-      const u = renderFlow.ui.action.split("/self-service/");
-      if (u.length == 2) {
-        renderFlow.ui.action = baseURL + "/api/kratos/self-service/" + u[1];
+      // call kratos through the go server proxy
+      const kratosUrl = renderFlow.ui.action.split("/self-service/");
+      if (kratosUrl.length == 2) {
```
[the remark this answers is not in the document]
edlerd: With the line above this ensures, that the ui action is a string that contains the path `/self-service/` exactly once. We only want to mess with the path in that case as that identifies, that we have a kratos api url at hand.

edlerd:
> Just one small query, also how do we go about testing this?

For testing [see this doc](https://docs.google.com/document/d/1Vb-ihoX_abfGWq5rnx9kHsE71YoFEI5h4XR6SgMHLFQ/edit?tab=t.0#heading=h.r81qbhxcvmff) with detailed descriptions.

mas-who (approved):

ui/pages/login.tsx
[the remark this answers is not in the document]
nsklikas: This would be gone if we used the sdk to somehow make the call here. Also ideally we wouldn't want to handle FORM requests in the backend. I will open an issue to properly investigate this and plan a way forward.

PR #498, opened 2025-02-20
Improve sequenced login design WD-19159

edlerd:
# Done
* add how to guide when setting up security key as part of a sequenced login
* Overhaul copy on sequenced sign in flow

### DIFF
--- a/ui/components/NodeInputButton.tsx
+++ b/ui/components/NodeInputButton.tsx
@@ -2,7 +2,6 @@ import { getNodeLabel } from "@ory/integrations/ui";
 import { Button } from "@canonical/react-components";
 import React, { FC, FormEvent } from "react";
 import { NodeInputProps } from "./helpers";
-import { ORY_LABEL_ID_ADD_SECURITY_KEY } from "../util/constants";
 import { isWebauthnAutologin } from "../util/webauthnAutoLogin";
 
 const getWebAuthnPayload = (evalCode: string): unknown => {
@@ -111,8 +110,16 @@ export const NodeInputButton: FC<NodeInputProps> = ({
   };
 
   // add security key button is supposed to be positive green
-  const appearance =
-    node.meta.label?.id === ORY_LABEL_ID_ADD_SECURITY_KEY ? "positive" : "";
+  let appearance = "";
+  let additional = null;
+  if (node.meta.label?.context) {
+    if ("appearance" in node.meta.label.context) {
+      appearance = node.meta.label.context.appearance as string;
+    }
+    if ("additional" in node.meta.label.context) {
+      additional = node.meta.label.context.additional;
+    }
+  }
 
   return (
     <>
@@ -124,6 +131,7 @@ export const NodeInputButton: FC<NodeInputProps> = ({
       >
         {getNodeLabel(node)}
       </Button>
+      {additional}
     </>
   );
 };
--- a/ui/pages/login.tsx
+++ b/ui/pages/login.tsx
@@ -333,7 +333,7 @@ const Login: NextPage = () => {
       });
     }
     if (isSignInWithHardwareKey(node) && isSequencedLogin) {
-      node.meta.label.text = "Sign in using Passkey";
+      node.meta.label.text = "Sign in using Security key";
     }
     return node;
   });
@@ -368,8 +368,7 @@ const Login: NextPage = () => {
         <>
           {isWebauthn && isSequencedLogin && (
             <p className="u-text--muted">
-              Another layer of authentication before you get access{" "}
-              {getTitleSuffix()}
+              Additional authentication needed to get access {getTitleSuffix()}
             </p>
           )}
           {flow ? (
--- a/ui/pages/setup_passkey.tsx
+++ b/ui/pages/setup_passkey.tsx
@@ -13,9 +13,10 @@ import { Flow } from "../components/Flow";
 import { kratos } from "../api/kratos";
 import PageLayout from "../components/PageLayout";
 import { AxiosError } from "axios";
-import { Button, Icon, Spinner } from "@canonical/react-components";
+import { Accordion, Button, Icon, Spinner } from "@canonical/react-components";
 import { UpdateSettingsFlowWithWebAuthnMethod } from "@ory/client/api";
 import {
+  ORY_LABEL_ID_ADD_SECURITY_KEY,
   ORY_LABEL_ID_NAME_OF_THE_SECURITY_KEY,
   ORY_LABEL_ID_REMOVE_SECURITY_ID,
 } from "../util/constants";
@@ -101,6 +102,9 @@ const SetupPasskey: NextPage = () => {
   const existingKeys: UiNode[] = [];
   const existingKeyNames: string[] = [];
 
+  const isSequencedFromSignInFlow =
+    flow?.return_to?.includes("/ui/login?") ?? false;
+
   const webauthnFlow = {
     ...renderFlow,
     ui: {
@@ -129,13 +133,72 @@ const SetupPasskey: NextPage = () => {
               deduplicateValues: existingKeyNames,
             };
           }
+          if (node.meta.label?.id === ORY_LABEL_ID_ADD_SECURITY_KEY) {
+            node.meta.label.context = {
+              ...node.meta.label.context,
+              appearance: isSequencedFromSignInFlow ? "" : "positive",
+              additional: (
+                <a
+                  className="p-button--link"
+                  href="mailto:support@canonical.com"
+                >
+                  Contact support
+                </a>
+              ),
+            };
+          }
           return node;
         }),
     },
   } as SettingsFlow;
 
   return (
-    <PageLayout title="Add a security key">
+    <PageLayout
+      title={isSequencedFromSignInFlow ? "Sign in" : "Add a security key"}
+    >
+      {isSequencedFromSignInFlow && (
+        <>
+          <p className="u-text--muted u-sv-3">
+            Authentication setup needed to continue.
+          </p>
+          <h2 className="p-heading--4">Add a security key</h2>
+          <Accordion
+            sections={[
+              {
+                title: "How to add a Security key",
+                content: (
+                  <>
+                    <ol className="p-list--nested-counter">
+                      <li>
+                        Enter a name for your security key (like {'"'}iPhone
+                        {'"'} or
+                        {'"'} Work Laptop{'"'})
+                      </li>
+                      <li>
+                        Click {'"'}Add security key{'"'}
+                      </li>
+
+                      <li>
+                        When your browser prompts you, use your device{"'"}s
+                        fingerprint, face recognition, or PIN to verify
+                      </li>
+                      <li>Wait for confirmation</li>
+                    </ol>
+                    <p>
+                      That{"'"}s it! Your passkey is now set up for future
+                      sign-ins.
+                    </p>
+                    <p>
+                      (Note: Your device needs biometrics or a PIN already
+                      configured)
+                    </p>
+                  </>
+                ),
+              },
+            ]}
+          />
+        </>
+      )}
       {webauthnFlow ? (
         <>
           <Flow onSubmit={handleSubmit} flow={webauthnFlow} />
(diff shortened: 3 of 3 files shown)


ui/pages/setup_passkey.tsx
```diff
@@ -202,6 +265,48 @@ const SetupPasskey: NextPage = () => {
...
           )}
+          {isSequencedFromSignInFlow && (
+            <>
+              <hr />
+              <div className="u-align--right">
+                <Button
+                  appearance="positive"
+                  type="button"
+                  className="u-no-margin--bottom u-align--right"
+                  onClick={() => {
```
BarcoMasile, line 276: inline func, please switch to either a func outside of the component or leverage `useCallback`

ui/components/NodeInputButton.tsx
```diff
@@ -111,8 +110,16 @@ export const NodeInputButton: FC<NodeInputProps> = ({
...
+  let appearance = "";
+  let additional = null;
+  if (node.meta.label?.context) {
+    if ("appearance" in node.meta.label.context) {
+      appearance = node.meta.label.context.appearance as string;
+    }
+    if ("additional" in node.meta.label.context) {
+      additional = node.meta.label.context.additional;
+    }
+  }
```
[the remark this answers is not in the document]
edlerd: We can probably split out the webauthn-login and webauthn-register buttons into their own component to increase cohesion. I'll put a story in next pulse to look into this.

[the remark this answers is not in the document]
edlerd: Created the [ticket](https://warthogs.atlassian.net/browse/WD-19549)

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+                </a>
+              ),
+            };
+          }
           return node;
         }),
     },
   } as SettingsFlow;
 
+  const highlightPasskeyRequirement = () => {
```
BarcoMasile, line 155: sorry to be pedantic, but even if you define the function as a constant, you still need `useCallback` to avoid instantiating it again at every render (which will keep the address changing)

I'll approve the PR to not block it for this, but consider using useCallback as the standard approach for these functions

BarcoMasile (dismissed):

nsklikas (commented):
Some minor questions. Also can you elaborate on:

> Missing a signal to show a confirmation screen after security key is added in sequenced login.

and 

> Missing a signal for the "Security key sign in failed" screen

Perhaps we can make them work, if required

ui/pages/login.tsx
```diff
@@ -366,8 +366,7 @@ const Login: NextPage = () => {
         <>
           {isWebauthn && isSequencedLogin && (
             <p className="u-text--muted">
-              Another layer of authentication before you get access{" "}
-              {getTitleSuffix()}
+              Additional authentication needed to get access {getTitleSuffix()}
```
natalian98, line 369: nitpick: Could you change the wording according to the designs?
```suggestion
              Additional authentication needed to access {getTitleSuffix()}
```

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+            sections={[
+              {
+                title: "How to add a Security key",
+                content: (
+                  <>
+                    <ol className="p-list--nested-counter">
+                      <li>
+                        Enter a name for your security key (like {'"'}iPhone
+                        {'"'} or
+                        {'"'} Work Laptop{'"'})
```
natalian98, line 200: nitpick: add blank space so that `or" Work Laptop"` becomes `or "Work Laptop"`
```suggestion
                        {'"'} or 
                        {'"'}Work Laptop{'"'})
```

natalian98 (commented):
Overall lgtm, thanks for the PR. I just have some minor comments.

Could you style this button so that it looks as in the figma screens?
image

===>

image

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+
   return (
-    <PageLayout title="Add a security key">
+    <PageLayout
+      title={isSequencedFromSignInFlow ? "Sign in" : "Add a security key"}
+    >
+      {isSequencedFromSignInFlow && (
+        <>
+          <p className="u-text--muted u-sv-3">
+            Authentication setup needed to continue.
```
natalian98, line 187:
```suggestion
            Authentication setup needed to access {getTitleSuffix()}
```

ui/pages/setup_passkey.tsx
```diff
@@ -202,6 +290,21 @@ const SetupPasskey: NextPage = () => {
...
+            <>
+              <hr />
+              <div className="u-align--right">
+                <Button
+                  appearance="positive"
+                  type="button"
+                  className="u-no-margin--bottom u-align--right"
+                  onClick={highlightPasskeyRequirement}
+                >
+                  Sign in
```
natalian98, line 303: Once you add a security key on the setup screen (first login), you are automatically signed in. I thought the purpose of this button is to let the user add more keys if needed and then click on `Sign in` when they're ready.

image

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
             };
           }
+          if (node.meta.label?.id === ORY_LABEL_ID_ADD_SECURITY_KEY) {
+            node.meta.label.context = {
+              ...node.meta.label.context,
+              appearance: isSequencedFromSignInFlow ? "" : "positive",
+              additional: (
+                <a
+                  className="p-button--link"
+                  href="mailto:support@canonical.com"
```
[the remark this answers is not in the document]
edlerd: Sure sounds all right. How can we expose the configuration for this and how can I load the configured value?

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+                      </li>
+
+                      <li>
+                        When your browser prompts you, use your device{"'"}s
+                        fingerprint, face recognition, or PIN to verify
+                      </li>
+                      <li>Wait for confirmation</li>
+                    </ol>
+                    <p>
+                      That{"'"}s it! Your passkey is now set up for future
```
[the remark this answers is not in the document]
edlerd: Right, this should probably be changed in the design as well. Below screen is from figma where I took this copy from.

FYI @lukasSerelis

image

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+    <PageLayout
+      title={isSequencedFromSignInFlow ? "Sign in" : "Add a security key"}
+    >
+      {isSequencedFromSignInFlow && (
+        <>
+          <p className="u-text--muted u-sv-3">
+            Authentication setup needed to continue.
+          </p>
+          <h2 className="p-heading--4">Add a security key</h2>
+          <Accordion
```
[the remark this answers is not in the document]
edlerd: Yes, makes sense to extract this into a dedicated component.

ui/pages/setup_passkey.tsx
```diff
@@ -129,13 +133,97 @@ const SetupPasskey: NextPage = () => {
...
+    if (!name.nextElementSibling) {
+      const warning = document.createElement("p");
+      warning.className = "p-form-validation__message";
+      warning.textContent = "You must set up a security key to sign in";
+      name.after(warning);
+    }
+
+    const groupSelector = ".p-form-validation";
+    const group = document.querySelectorAll(groupSelector);
+    group.forEach((e) => e.classList.add("is-error"));
```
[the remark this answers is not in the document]
edlerd: I agree, ideally we don't need plain JS here. We have no way to reference the nested UI flow object that kratos gives us, which is rendering the input field from here. So this is the easiest way I could think of to connect the "sign in" button with that. We can extract all of this code into a dedicated component though to make it a bit cleaner in here.

ui/pages/login.tsx
edlerd: I wanted to reuse the same function for the title in here. That's why I slightly adjusted the copy so it fits. I think it should be fine though. Both copies seem equivalent to me. Wdyt?

ui/pages/setup_passkey.tsx
edlerd: We don't have the title suffix in here. See my comments on the description above about missing signals and adjusting the design to make fit for that.

ui/pages/setup_passkey.tsx
edlerd: When we discussed it with Lukas he mentioned this button is here to let the users think they are still in the sign in flow. See my comments in the description above about the missing signal and the compromise to not allow adding multiple keys as the flow finishes automatically, and we have no way to intercept it in the frontend. If you see a way to get that signal I am happy to use it.

ui/pages/login.tsx
edlerd: The underlying problem this is aiming to solve is that getTitleSuffix can return "to foo" and "foo" depending on other circumstances. Hence the idea to find a copy here that is fit for both.

ui/pages/setup_passkey.tsx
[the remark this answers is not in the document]
edlerd: From what I understood discussing this with @lukasSerelis it is a mere placeholder at this point. Maybe we should remove this completely until and agreement is reached?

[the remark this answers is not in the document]
nsklikas: Sounds good to me. Could probably reuse the `app-config` API to provide such configuration to the frontend

nsklikas (approved):
Approving, feel free to either remove or keep the mailto `href` as it is, let's just open an issue to track it

edlerd:
> Overall lgtm, thanks for the PR. I just have some minor comments.
> 
> Could you style this button so that it looks as in the figma screens? image
> 
> ===>
> 
> image

I guess you are talking about the icon, right? I added it just now. Btw. please always comment on a file or line, otherwise GitHub won't allow to have proper threading on those discussions :)

edlerd:
I think all comments above are resolved or at least responded. Please let me know if you have further suggestions :smiley_cat:

ui/pages/setup_passkey.tsx
natalian98, line 1: > I guess you are talking about the icon, right? I added it just now.

@edlerd I meant the icon and styling of the button, both when signing in and on the setup screen. Right now the buttons are grey (I think it's a default style from kratos?), designs suggest white. It's a minor thing but let's try to follow those.

ui/pages/setup_passkey.tsx
natalian98: It was related to my [comment](https://www.figma.com/design/zstuCMdHsKx8Wr9ERZx9Kw?node-id=630-24118#1116366810) in figma, if you'd be able to tie the `return_to` with the sign in button.
Feel free to disregard if that was agreed with Lukas.

edlerd: On adding a security key, the form is submitting to kratos with a post request (see below). The response has a redirect header to the login form. I don't see a way to intercept that. On the login page, I also don't see a way to identify the user is just adding their first key in order to send them back to the setup passkey page. Do you see a way to do that @natalian98 ?

image

natalian98: I was asking about that because the settings form is sent by the webauthn script (see https://github.com/canonical/identity-platform-login-ui/blob/955e47f3b6269de9170aba91c65c8fdf28c99eb1/ui/pages/setup_passkey.tsx#L77). If you don't see a way to intercept it then let's leave it.

natalian98 (approved):

BarcoMasile (approved):

mas-who (approved):

lukasSerelis:
Looks good! The generic Sign in to "OIDC App" is a bit sad, could we replace the helper text to say "Additional authentication needed to get access to this Application" instead? Same for the header, just saying "Sign in" is likely going to cause less confusion.

image

lukasSerelis:
> > Looks good! The generic Sign in to "OIDC App" is a bit sad, could we replace the helper text to say "Additional authentication needed to get access to this Application" instead? Same for the header, just saying "Sign in" is likely going to cause less confusion.
> 
> I agree it looks wired in this example, but assuming it is due to the odd name of the "OIDC App". In practice, this will say something like "Sign in to juju.canonical.com" or "Sign in to grafana". The words after "to" are dynamic and based on the app name or domain.
 
 Understood. Then it all looks good!

PR #513, opened 2025-02-26
fix(reset) return to login after resetting password

### LINKED ISSUE #366: User not redirected back to app
On the e2e tests if the user tries to log in, but decides to reset their password, they do not get redirected back on the original application.

Is this the intended behavior?

(see https://github.com/canonical/identity-platform-login-ui/pull/365#discussion_r1850700955)

edlerd:
# Done
- include return to when starting reset password flow from login

fixes #366


nsklikas (commented):
Can you add an e2e test for this? Uncommenting https://github.com/canonical/identity-platform-login-ui/blob/main/ui/tests/reset-password.spec.ts#L29 should be enough.

edlerd:
> Can you add an e2e test for this? Uncommenting https://github.com/canonical/identity-platform-login-ui/blob/main/ui/tests/reset-password.spec.ts#L29 should be enough.

Good hint, thanks! I updated the test.

How nice of the e2e test suite: It surfaced a missing piece, that is now fixed. I also updated one of the screenshots from the reset password flow, because indeed that step changed :)

nsklikas (approved):
Nice!! LGTM

natalian98 (approved):
lgtm, thanks for the fix
Could you rebase the PR?

edlerd:
Thanks for the reviews @natalian98 and @nsklikas 
Just pushed the rebase on main.

PR #560, opened 2025-04-01
Add self serve

edlerd:
# Done
* update ui packages
* add self serve area


ui/pages/self_serve_details.tsx
```diff
@@ -0,0 +1,75 @@
...
+            "",
+            `./self_serve_details?flow=${data.id}`,
+          );
+          router.query.flow = data.id;
+        }
+        setFlow(data);
+      })
+      .catch(handleFlowError("settings", setFlow))
+      .catch(async (err: AxiosError<string>) => {
+        if (err.response?.data.trim() === "Failed to create settings flow") {
```
BarcoMasile, line 54: question: do we have to rely on this text? Shouldn't there be an error ID here? I may remember wrong :-/

BarcoMasile (changes requested):
Seems LGTM, I just need those react hooks to avoid rerenders

ui/util/selfServeHelpers.tsx
```diff
@@ -0,0 +1,50 @@
+import { SettingsFlow } from "@ory/client";
+
+export const formatReturnTo = (
+  returnTo: string | string[] | undefined,
+  forceSelfServe?: boolean,
+) => {
+  if (forceSelfServe) {
+    const origin = window.location.origin;
+    const path = window.location.pathname;
+    const basePath = path.substring(0, path.lastIndexOf("/"));
```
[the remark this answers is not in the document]
edlerd: All the login UI URLs are of the form `.../ui/page`. We strip the `/page` from the current URL in this function to produce the base route independent of the prefix for the deployment.

ui/pages/self_serve_details.tsx
edlerd: We can see in kratos/handlers.go that here is no other information available. Just an http status 500, but that would be too generic.

BarcoMasile (dismissed):
Thanks, approved. I would suggest to wait until our meeting tomorrow before merging if that's ok

lukasSerelis:
This all looks great, we will however need to decide on how users get to this page, how do admins set up these self service pages, mostly future concerns, but this is a great piece of work @edlerd

ui/util/selfServeHelpers.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  if (traits.name && traits.surname) {
+    return traits.name + " " + traits.surname;
+  }
+
+  if (traits.identity) {
+    return traits.identity;
+  }
+
+  return "";
```
anusha-c18, line 44: Should we return the email as the last resort if no other value is defined?

edlerd: Good idea for the side navigation. We should distinguish the full name for the manage detail page. I'll split off another function for that and make the user name input field for full name render conditionally.

ui/components/SelfServeNavigation.tsx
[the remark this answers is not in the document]
edlerd: Yes, good idea!

anusha-c18 (dismissed):
LGTM 👍

BarcoMasile (approved):

nsklikas (commented):
Overall, lgtm. A couple of comments:
- Should we restrict this page from oidc users? Do we want to allow them to set the password?
- If the name is too long then it cuts out (see the bottom left of the picture
image

I think that my 1st comment should be part of a broad discussion (related to account linking), so it should not affect this PR.

edlerd:
Good point @nsklikas 

I added truncation and a title, so on hover the full context can be discovered.

nsklikas (approved):
lgtm

BarcoMasile (approved):

anusha-c18 (approved):
LGTM 👍

Issue #374, opened 2024-11-26
Remove MethodOnly struct


nsklikas:
On https://github.com/canonical/identity-platform-login-ui/blob/main/pkg/kratos/service.go#L585 we have a comment about refactoring once we update the sdk version. Is that still valid? Should we open an issue about it or the problem was not solved?

_Originally posted by @nsklikas in https://github.com/canonical/identity-platform-login-ui/pull/372#pullrequestreview-2461495884_

PR #577, opened 2025-04-14
fix: clean the request body parsing in login & settings flow

wood-push-melon:
This pull request aims to resolve the https://github.com/canonical/identity-platform-login-ui/issues/374. However, I decided to keep using the custom one, and not use the `SessionAuthenticationMethod` in the krato's SDK for the following reasons (Let me know if I'm off the road):

- It's not necessary because we only care about the authentication method that the user selects in the flow.
- Based on what I understand in the SDK's source code, the `SessionAuthenticationMethod` is to provide information about the authentication method that **_WAS_** used in the authenticated session.

### DIFF
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -596,114 +596,87 @@ func (s *Service) FilterFlowProviderList(ctx context.Context, flow *kClient.Logi
 }
 
 func (s *Service) ParseLoginFlowMethodBody(r *http.Request) (*kClient.UpdateLoginFlowBody, []*http.Cookie, error) {
-	// TODO: try to refactor when we bump kratos sdk to 1.x.x
-	var (
-		ret     kClient.UpdateLoginFlowBody
-		cookies = r.Cookies()
-	)
-	methodOnly := new(methodOnly)
-
 	defer r.Body.Close()
-	b, err := io.ReadAll(r.Body)
 
+	cookies := r.Cookies()
+	bodyBytes, err := io.ReadAll(r.Body)
 	if err != nil {
 		return nil, cookies, errors.New("unable to read body")
 	}
 
 	// replace the body that was consumed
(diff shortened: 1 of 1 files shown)


pkg/kratos/service.go
```diff
@@ -596,114 +596,87 @@ func (s *Service) FilterFlowProviderList(ctx context.Context, flow *kClient.Logi
...
-			if c.Name == KRATOS_SESSION_COOKIE_NAME {
-				if i == len(cookies)-1 {
-					cookies = cookies[:i]
-				} else {
-					cookies[i] = cookies[len(cookies)-1]
-					cookies = cookies[:len(cookies)-1]
-				}
+	// Remove session cookie if this is a 1FA method
+	if s.is1FAMethod(methodPayload.Method) {
+		filtered := cookies[:0]
```
BarcoMasile, line 673: nitpick: I think this is ok, but just to stress the inner workings of the go language: this statement here doesn't actually instantiate a new slice (like in python for example). What happens is that `filtered` will hold a reference to the original `cookies`, and length and capacity properties will be modified by the `append` operation.
This means that the original cookies slice values will be replaced from the beginning with the kratos session cookie from the head of the list, but old values will still be there. Since len and cap would be updated accordingly this should not be a problem.
But, it could be a little cleaner to just instantiate a new completely new slice with

```go
  filtered := make([]*http.Cookie, 0)

```

BarcoMasile (approved):
LGTM

nsklikas (dismissed):
lgtm, I agree with Marco

pkg/kratos/service.go
wood-push-melon: yep, that's what I learned from this [source](https://josh-weston.medium.com/golang-in-place-slice-operations-5607fd90217). But happy to update it using the new slice to make it clear.

BarcoMasile (approved):

PR #677, opened 2025-08-06
Identifier first pattern support

natalian98:
This PR adds backend support for identifier first login pattern.
- adds `/self-service/login/id-first` endpoint to handle identifier first login flow
- adds `IDENTIFIER_FIRST_ENABLED` environment variable and expose it in `app-config` endpoint
- bypass kratos SDK for updating login and make a low-level http call to kratos with `Content-Type: application/x-www-form-urlencoded` to avoid hitting [this](https://github.com/ory/kratos/blob/15f3eb3aa25c92e3d1e025082226f26d2ab58718/selfservice/strategy/idfirst/strategy_login.go#L150C2-L154C3) error

### DIFF
--- a/internal/kratos/client_test.go
+++ b/internal/kratos/client_test.go
@@ -1,0 +1,35 @@
+package kratos
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestExecuteIdentifierFirstUpdateLoginRequest(t *testing.T) {
+	cookies := make([]*http.Cookie, 0)
+	cookie := &http.Cookie{Name: "test", Value: "test"}
+	cookies = append(cookies, cookie)
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		if req.Method != http.MethodPost {
+			t.Fatalf("expected method POST, got %s", req.Method)
+		}
+		http.SetCookie(w, cookie)
+		w.WriteHeader(http.StatusSeeOther)
+	}))
+	defer server.Close()
+
+	client := NewClient(server.URL, false)
+	ctx := context.Background()
+
+	resp, err := client.ExecuteIdentifierFirstUpdateLoginRequest(ctx, "flow123", "csrf_token_1234", "test@example.com", cookies)
+
+	if err != nil {
+		t.Fatalf("expected error to be nil not %v", err)
+	}
+	if resp.StatusCode != http.StatusSeeOther {
+		t.Errorf("expected status 303, got %d", resp.StatusCode)
+	}
+}
--- a/pkg/kratos/handlers_test.go
+++ b/pkg/kratos/handlers_test.go
@@ -20,17 +20,18 @@ import (
 )
 
 const (
-	BASE_URL                        = "https://example.com"
-	HANDLE_CREATE_FLOW_URL          = BASE_URL + "/api/kratos/self-service/login/browser"
-	HANDLE_UPDATE_LOGIN_FLOW_URL    = BASE_URL + "/api/kratos/self-service/login"
-	HANDLE_GET_LOGIN_FLOW_URL       = BASE_URL + "/api/kratos/self-service/login/flows"
-	HANDLE_ERROR_URL                = BASE_URL + "/api/kratos/self-service/errors"
-	HANDLE_CREATE_RECOVERY_FLOW_URL = BASE_URL + "/api/kratos/self-service/recovery/browser"
-	HANDLE_UPDATE_RECOVERY_FLOW_URL = BASE_URL + "/api/kratos/self-service/recovery"
-	HANDLE_GET_RECOVERY_FLOW_URL    = BASE_URL + "/api/kratos/self-service/recovery/flows"
-	HANDLE_CREATE_SETTINGS_FLOW_URL = BASE_URL + "/api/kratos/self-service/settings/browser"
-	HANDLE_UPDATE_SETTINGS_FLOW_URL = BASE_URL + "/api/kratos/self-service/settings"
-	HANDLE_GET_SETTINGS_FLOW_URL    = BASE_URL + "/api/kratos/self-service/settings/flows"
+	BASE_URL                                      = "https://example.com"
+	HANDLE_CREATE_FLOW_URL                        = BASE_URL + "/api/kratos/self-service/login/browser"
+	HANDLE_UPDATE_LOGIN_FLOW_URL                  = BASE_URL + "/api/kratos/self-service/login"
+	HANDLE_UPDATE_IDENTIFIER_FIRST_LOGIN_FLOW_URL = BASE_URL + "/api/kratos/self-service/login/id-first"
+	HANDLE_GET_LOGIN_FLOW_URL                     = BASE_URL + "/api/kratos/self-service/login/flows"
+	HANDLE_ERROR_URL                              = BASE_URL + "/api/kratos/self-service/errors"
+	HANDLE_CREATE_RECOVERY_FLOW_URL               = BASE_URL + "/api/kratos/self-service/recovery/browser"
+	HANDLE_UPDATE_RECOVERY_FLOW_URL               = BASE_URL + "/api/kratos/self-service/recovery"
+	HANDLE_GET_RECOVERY_FLOW_URL                  = BASE_URL + "/api/kratos/self-service/recovery/flows"
+	HANDLE_CREATE_SETTINGS_FLOW_URL               = BASE_URL + "/api/kratos/self-service/settings/browser"
+	HANDLE_UPDATE_SETTINGS_FLOW_URL               = BASE_URL + "/api/kratos/self-service/settings"
+	HANDLE_GET_SETTINGS_FLOW_URL                  = BASE_URL + "/api/kratos/self-service/settings/flows"
 )
 
 //go:generate mockgen -build_flags=--mod=mod -package kratos -destination ./mock_logger.go -source=../../internal/logging/interfaces.go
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -63,6 +63,7 @@ At the moment the application is sourcing the following from the environment:
 - `OPENFGA_MODEL_ID` - the OpenFGA model ID to use. If not specified, a new
   model will be created
 - `MFA_ENABLED` - whether MFA is enabled and enforced, defaults to true
+- `IDENTIFIER_FIRST_ENABLED` - whether login flow follows the identifier-first pattern, defaults to true
```
natalian98, line 66: This should be true by default if we follow the design

pkg/kratos/service.go
```diff
@@ -387,6 +391,74 @@ func (s *Service) UpdateRecoveryFlow(
...
+
+	form := url.Values{}
+	form.Set("csrf_token", *csrfToken)
+	form.Set("identifier", *identifier)
+	form.Set("method", "identifier_first")
+
+	u, err := url.Parse(s.kratosPublicURL + "/self-service/login")
+	if err != nil {
+		return nil, nil, fmt.Errorf("invalid kratos URL: %w", err)
+	}
```
BarcoMasile, line 418: this doesn't change for every invocation, it could be computed once and reused

pkg/kratos/service.go
```diff
@@ -387,6 +391,74 @@ func (s *Service) UpdateRecoveryFlow(
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), strings.NewReader(form.Encode()))
+	if err != nil {
+		s.logger.Errorf("failed to create http request: %s", err)
+		return nil, nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	for _, c := range cookies {
+		req.AddCookie(c)
+	}
```
BarcoMasile, line 431: please isolate this "request building" logic in a separate private function

in alternative, you could unify the request building logic with the request invocation that happens at line 441

pkg/kratos/service.go
```diff
@@ -387,6 +391,74 @@ func (s *Service) UpdateRecoveryFlow(
...
+
+	client := &http.Client{
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			// Kratos returns 303 for identifier first flows. We disable these automatic redirects
+			// in order to let the handler process the BrowserLocationChangeRequired and change to 200
+			return http.ErrUseLastResponse
+		},
+	}
```
BarcoMasile: issue: this should be reused a single instance, and instantiating every time may cause issue since we would not have any control over how many connections are established with kratos

pkg/kratos/service.go
```diff
@@ -942,14 +1024,16 @@ func (s *Service) parseKratosRedirectResponse(ctx context.Context, resp *http.Re
 		Error:      redirectResp.Error,
 	}, nil
 }
-func NewService(kratos KratosClientInterface, kratosAdmin KratosAdminClientInterface, hydra HydraClientInterface, authzClient AuthorizerInterface, oidcWebAuthnSequencingEnabled bool, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {
+func NewService(kratos KratosClientInterface, kratosAdmin KratosAdminClientInterface, hydra HydraClientInterface, authzClient AuthorizerInterface, KratosPublicURL string, oidcWebAuthnSequencingEnabled bool, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {
```
BarcoMasile, line 1025: nitpick: KratosPublicURL parameter should be lowercase
```suggestion
func NewService(kratos KratosClientInterface, kratosAdmin KratosAdminClientInterface, hydra HydraClientInterface, authzClient AuthorizerInterface, kratosPublicURL string, oidcWebAuthnSequencingEnabled bool, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {

```

pkg/kratos/service.go
```diff
@@ -387,6 +391,72 @@ func (s *Service) UpdateRecoveryFlow(
...
+	if !ok {
+		return nil, nil, fmt.Errorf("missing identifier")
+	}
+
+	form := url.Values{}
+	form.Set("csrf_token", *csrfToken)
+	form.Set("identifier", *identifier)
+	form.Set("method", "identifier_first")
+
+	u, err := url.Parse(s.kratosPublicURL + "/self-service/login")
```
wood-push-melon, line 415: (feel free to ignore)

It seems that this logic can be extracted out of the method. Or maybe put it at the beginning of the method (if it has to stay in this method).

pkg/kratos/service.go
```diff
@@ -491,6 +561,8 @@ func (s *Service) getUiError(responseBody io.ReadCloser) (err error) {
 	switch errorCode := errorCodes[0].Id; errorCode {
 	case IncorrectCredentials:
 		err = fmt.Errorf("incorrect username or password")
+	case IncorrectAccountIdentifier:
+		err = fmt.Errorf("account does not exist or has no login method configured")
```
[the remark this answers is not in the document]
natalian98: That's correct - they must either register with some IdP first or link it to an account to have it listed in the second login step

pkg/kratos/service.go
```diff
@@ -387,6 +391,72 @@ func (s *Service) UpdateRecoveryFlow(
...
+	q := u.Query()
+	q.Set("flow", flow)
+	u.RawQuery = q.Encode()
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u.String(), strings.NewReader(form.Encode()))
+	if err != nil {
+		s.logger.Errorf("failed to create http request: %s", err)
+		return nil, nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
```
shipperizer, line 428: https://pkg.go.dev/github.com/ory/kratos-client-go#LoginFlow.SetType

based on what I read, if we set the type to API we won't need to perform form posts anymore, same applies from the frontend potentially

same argument is probably valid for the passkey setup

cmd/serve.go
```diff
@@ -93,7 +93,7 @@ func serve() {
 		panic("Invalid authorization model provided")
 	}
 
-	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.OIDCWebAuthnSequencingEnabled, specs.BaseURL, specs.SupportEmail, specs.KratosPublicURL, tracer, monitor, logger)
+	router := web.NewRouter(kClient, kAdminClient, hClient, authorizer, cookieManager, distFS, specs.MFAEnabled, specs.OIDCWebAuthnSequencingEnabled, specs.IdentifierFirstEnabled, specs.BaseURL, specs.SupportEmail, specs.KratosPublicURL, tracer, monitor, logger)
```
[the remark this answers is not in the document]
BarcoMasile: https://github.com/canonical/identity-platform-login-ui/issues/685

pkg/kratos/service.go
```diff
@@ -387,6 +393,78 @@ func (s *Service) UpdateRecoveryFlow(
...
+
+	resp, err := s.executeIdentifierFirstUpdateLoginRequest(ctx, flow, *csrfToken, *identifier, cookies)
+	if err != nil {
+		s.logger.Errorf("kratos request failed: %s", err)
+		return nil, nil, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusSeeOther {
```
nsklikas, line 455: nit: happy path should not be in if or maybe a switch would be more appropriate.

pkg/kratos/service.go
```diff
@@ -387,6 +393,78 @@ func (s *Service) UpdateRecoveryFlow(
...
+		location := resp.Header.Get("Location")
+		return &BrowserLocationChangeRequired{RedirectTo: &location}, resp.Cookies(), nil
+	}
+
+	if resp.StatusCode == http.StatusBadRequest {
+		err = s.getUiError(resp.Body)
+		return nil, nil, err
+	}
+
+	return nil, nil, fmt.Errorf("unexpected status: %d", resp.StatusCode)
```
nsklikas, line 465: suggestion: log something here, not sure when this can happen but we should have some info about it

pkg/kratos/service.go
```diff
@@ -387,6 +393,78 @@ func (s *Service) UpdateRecoveryFlow(
 	return returnToResp, resp.Cookies(), nil
 }
 
+func (s *Service) executeIdentifierFirstUpdateLoginRequest(
```
nsklikas, line 396: thought: this looks like logic that should be part of the internal kratos client, not service logic

nsklikas, line 396: > bypass kratos SDK for updating login and make a low-level http call to kratos with Content-Type: application/x-www-form-urlencoded to avoid hitting [this](https://github.com/ory/kratos/blob/15f3eb3aa25c92e3d1e025082226f26d2ab58718/selfservice/strategy/idfirst/strategy_login.go#L150C2-L154C3) error

question: looking at that sample of code and [this](https://github.com/ory/herodot/blob/master/json.go#L100) function, it looks like even though the response is an error the flow is in there. Can't we use the sdk and get the flow from the response body?

nsklikas (changes requested):
couple of nits, I also have a question about whether we have to bypass the sdk

pkg/kratos/service.go
natalian98: Sure:
1. Clone `IAM-1639-identifier-first-pattern` branch and apply [this](https://gist.github.com/natalian98/c35da57481427b652f866094802795d6) patch for frontend on top.
2. In `kratos.yaml`, switch the login style to `identifier_first`.
3. Build and run login ui and the dev docker compose as usual.
4. Go to login page, it will be divided into 2 steps: in first you provide email, next you'll be shown available login methods.

If you want to test calling sdk, just switch line 68 in the patch file to `/api/kratos/self-service/login?flow=${encodeURIComponent(flowId)}`

pkg/kratos/service.go
natalian98: changed to a switch

pkg/kratos/service.go
natalian98: As discussed privately it appears to be a debugger issue as I can also get the flow from logs.
That said, 400 means form validation errors, causing the login flow update to fail. The returned flow object reflects this failure, so I still believe using the SDK isn’t feasible for this method. In a happy path, we should get a 303 and redirect to the next step, just like with my workaround.

nsklikas: 400 means that the server considers that there was a client error, in this case this means that a wrong content type was requested, not sure if this is a form validation error. IMO you should not get a 303 in a backchannel request, as a backend would not want to be served an html page. When I manually navigate my browser to `<login_ui_url>/ui/login?flow=<flow_id>` (after the sdk call), I can see that the flow has been updated.

I can see the argument that we should not expect an error response in a non-error path, but I think we should consider it since we wouldn't have to implement our own http logic.

pkg/kratos/service.go
BarcoMasile: put another comment on this since it has not been addressed yet, closing this in favor of the other comment

pkg/kratos/service.go
natalian98: I posted a question on the ory slack about the 400, let's wait for their reply and discuss next week @shipperizer @BarcoMasile

BarcoMasile: my 2 cents here: relying on a 400 error to handle a successful case is too fragile an approach and error prone. Also, we cannot expect Ory to keep a "successful" behavior when an error code is returned, since we would not have any guarantee that that behavior is kept in the future.
If I understand this thread correctly, then I suggest we don't rely on an error for our case.

internal/kratos/client.go
```diff
@@ -23,20 +33,73 @@ func (c *Client) MetadataApi() client.MetadataAPI {
...
+	for _, cookie := range cookies {
+		req.AddCookie(cookie)
+	}
+
+	return req, nil
+}
+
+func (c *Client) ExecuteIdentifierFirstUpdateLoginRequest(
+	ctx context.Context, flow string, csrfToken, identifier string, cookies []*http.Cookie,
+) (*http.Response, error) {
```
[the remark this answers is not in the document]
nsklikas: I agree it isn't necessary, just wanted to add to have a consistent interface as all other methods have this signature. Feel free to do what you want

pkg/kratos/service.go
nsklikas: Great and we already have a function for parsing the flow in case of error. Why don't we parse the flow in order to decide whether an error actually occurred or not?

internal/kratos/client.go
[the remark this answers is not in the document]
shipperizer: let's open an issue about it under the `enhancement` label, then let's move on

internal/kratos/client.go
```diff
@@ -23,20 +33,69 @@ func (c *Client) MetadataApi() client.MetadataAPI {
...
 	}
 
-	configuration.HTTPClient = &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+	httpClient := &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+	configuration.HTTPClient = httpClient
 
-	c.c = client.NewAPIClient(configuration)
+	loginURL, err := url.Parse(baseURL + "/self-service/login")
+	if err != nil {
+		panic(fmt.Sprintf("invalid kratos login url: %v", err))
```
shipperizer, line 93: no panic guys, unless we add a recover please

idea is to return the error and fail in the main or wherever we instantiate it

pkg/kratos/service.go
shipperizer: as mentioned in https://github.com/canonical/identity-platform-login-ui/pull/697#discussion_r2272479968 let's tackle this in a separate issue aimed at refactoring the interaction between kratos, backend and frontend.

let's move on from this

cc @nsklikas @natalian98

BarcoMasile (approved):

wood-push-melon (dismissed):

internal/kratos/client.go
shipperizer: my bad, didn't realize the NewClient doesn't return an error together with the pointer.

let's stick with panic for now

nsklikas (dismissed):

Issue #305, opened 2024-10-07
Make base path dynamic, same way as it's done for Admin UI in #426


BarcoMasile:
## Description
Perform on Login UI same improvement from https://github.com/canonical/identity-platform-admin-ui/pull/426 in Admin UI to allow for a dynamic base path through templating on the index.html

FYI @nsklikas @edlerd

PR #697, opened 2025-08-12
IAM 1403

shipperizer:
- **ci: add skaffold to build and push**
- **fix: switch to basic loading of webauthn script**
- **fix: ui can only talk to login svc backend**
- **fix: if request expects a non json response, then redirect to ui**
- **fix: dont clear cookies if we are not in an aal2 state or if totp or webauthn are disabled**
- **fix: deal with settings flow while setting up the passkeys, accept form data and redirect like kratos does**

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -60,6 +60,13 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 
 	q := r.URL.Query()
 
+	if q.Get("aal") == "aal2" && r.Header.Get("Accept") != "application/json, text/plain, */*" {
+		u, _ := url.JoinPath(a.baseURL, "/ui/login")
+		u = u + "?" + r.URL.RawQuery
+		http.Redirect(w, r, u, http.StatusSeeOther)
+		return
+	}
+
 	loginChallenge := q.Get("login_challenge")
 	returnTo := q.Get("return_to")
 	aal := q.Get("aal")
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -149,6 +149,7 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	}
 
 	setCookies(w, cookies)
+
 	w.WriteHeader(http.StatusOK)
 	_ = json.NewEncoder(w).Encode(response)
 }
@@ -156,13 +157,21 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 func (a *API) handleCreateFlowNewSession(r *http.Request, aal string, returnTo string, loginChallenge string, refresh bool) (*client.LoginFlow, []*http.Cookie, error) {
 	// redirect user to this endpoint with the login_challenge after login
 	// see https://github.com/ory/kratos/issues/3052
+
+	cookies := r.Cookies()
+
+	// if not aal2 or both mfa and webauthn disabled clear cookies
+	if aal != "aal2" || !(a.mfaEnabled || a.oidcWebAuthnSequencingEnabled) {
+		cookies = filterCookies(cookies, KRATOS_SESSION_COOKIE_NAME)
+	}
+
 	flow, cookies, err := a.service.CreateBrowserLoginFlow(
 		r.Context(),
 		aal,
 		returnTo,
 		loginChallenge,
 		refresh,
-		filterCookies(r.Cookies(), KRATOS_SESSION_COOKIE_NAME),
+		cookies,
 	)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to create login flow, err: %v", err)
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -731,15 +731,14 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	resp, err := json.Marshal(flow)
-	if err != nil {
-		a.logger.Errorf("Error when marshalling json: %v\n", err)
-		http.Error(w, "Failed to parse settings flow", http.StatusInternalServerError)
+	// maintain previous kratos behaviour
+	setCookies(w, cookies)
+	if returnTo, ok := flow.GetReturnToOk(); ok {
+		a.redirectResponse(w, r, &BrowserLocationChangeRequired{RedirectTo: returnTo})
 		return
 	}
-	setCookies(w, cookies)
-	w.WriteHeader(http.StatusOK)
-	w.Write(resp)
+
+	http.Error(w, "unexpected error", http.StatusInternalServerError)
 }
 
 func (a *API) handleCreateSettingsFlow(w http.ResponseWriter, r *http.Request) {
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io/fs"
 	"net/http"
-	"net/url"
 	"path"
 	"strings"
 
@@ -68,15 +67,8 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
 	a.fileServer.ServeHTTP(w, r)
 }
 
-func (a *API) getCSP(baseURL, kratosPublicURL string) string {
-	b, _ := url.Parse(baseURL)
-	k, _ := url.Parse(kratosPublicURL)
-	additionalScriptURL := ""
-	if k != nil && b != nil && k.Host != b.Host {
-		// Allowlist the kratos URL to allow the browser needs to fetch the webauthn.js script
-		additionalScriptURL = kratosPublicURL
-	}
-	return fmt.Sprintf("default-src 'self' data: https://assets.ubuntu.com; script-src 'self' %v; style-src 'self'", additionalScriptURL)
+func (a *API) getCSP(baseURL string) string {
+	return fmt.Sprintf("default-src 'self' data: https://assets.ubuntu.com; script-src 'self' %v/.well-known/webauthn.js; style-src 'self'", baseURL)
 }
 
 func NewAPI(fileSystem fs.FS, baseURL, kratosPublicURL string, logger logging.LoggerInterface) *API {
@@ -86,7 +78,7 @@ func NewAPI(fileSystem fs.FS, baseURL, kratosPublicURL string, logger logging.Lo
 
 	a.baseURL = baseURL
 	a.kratosPublicURL = kratosPublicURL
-	a.csp = a.getCSP(baseURL, kratosPublicURL)
+	a.csp = a.getCSP(baseURL)
 	a.logger = logger
 
 	return a
--- a/ui/components/Node.tsx
+++ b/ui/components/Node.tsx
@@ -3,13 +3,13 @@ import {
   isUiNodeAnchorAttributes,
   isUiNodeImageAttributes,
   isUiNodeInputAttributes,
-  isUiNodeScriptAttributes,
+  // isUiNodeScriptAttributes,
   isUiNodeTextAttributes,
 } from "@ory/integrations/ui";
 import { NodeAnchor } from "./NodeAnchor";
 import { NodeImage } from "./NodeImage";
 import { NodeInput } from "./NodeInput";
-import { NodeScript } from "./NodeScript";
+// import { NodeScript } from "./NodeScript";
 import { NodeText } from "./NodeText";
 import { NodeInputProps } from "./helpers";
 import { FC } from "react";
@@ -26,9 +26,9 @@ export const Node: FC<Omit<NodeInputProps, "attributes">> = ({
     return <NodeImage node={node} attributes={node.attributes} />;
   }
 
-  if (isUiNodeScriptAttributes(node.attributes)) {
-    return <NodeScript node={node} attributes={node.attributes} />;
-  }
+  // if (isUiNodeScriptAttributes(node.attributes)) {
+  //   return <NodeScript node={node} attributes={node.attributes} />;
+  // }
 
   if (isUiNodeTextAttributes(node.attributes)) {
     return <NodeText node={node} attributes={node.attributes} />;
--- a/ui/components/PageLayout.tsx
+++ b/ui/components/PageLayout.tsx
@@ -27,6 +27,7 @@ const PageLayout: FC<Props> = ({ children, title, user, isSelfServe }) => {
           href="https://assets.ubuntu.com/v1/16c27f81-COF%20favicon-16x16.png"
         />
         <title>{title}</title>
+        <script src="../.well-known/webauthn.js" />
       </Head>
       {isSelfServe ? (
         <div className="l-application" role="presentation">
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# The script requires:
+# - rockcraft
+# - skopeo with sudo privilege
+# - yq
+# - docker
+
+set -e
+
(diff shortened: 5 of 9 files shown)


pkg/kratos/handlers.go
```diff
@@ -60,6 +60,13 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 
 	q := r.URL.Query()
 
+	if q.Get("aal") == "aal2" && r.Header.Get("Accept") != "application/json, text/plain, */*" {
```
nsklikas, line 63: Similaril to the discussion in https://github.com/canonical/identity-platform-login-ui/pull/677#discussion_r2266762829.

The reason that we need the call to kratos to be through a backchannel call is that when using the Kratos sdk we always get back a JSON object, and we can't serve the browser a JSON object when it requires html. Another approach to overcome this issue would be to keep the current logic and after we create the flow to just redirect the user to the login page with the flow ID. A dirty way to do that would be:
```
diff --git a/pkg/kratos/handlers.go b/pkg/kratos/handlers.go
index caadeb4..404b472 100644
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -52,6 +52,7 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
        var (
                response              any
+               responseFlow          *client.LoginFlow
                shouldEnforceMfa      = false
                shouldEnforceWebAuthn = false
                cookies               []*http.Cookie
@@ -60,12 +61,12 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 
        q := r.URL.Query()
 
-       if q.Get("aal") == "aal2" && r.Header.Get("Accept") != "application/json, text/plain, */*" {
-               u, _ := url.JoinPath(a.baseURL, "/ui/login")
-               u = u + "?" + r.URL.RawQuery
-               http.Redirect(w, r, u, http.StatusSeeOther)
-               return
-       }
+       // if q.Get("aal") == "aal2" && r.Header.Get("Accept") != "application/json, text/plain, */*" {
+       //      u, _ := url.JoinPath(a.baseURL, "/ui/login")
+       //      u = u + "?" + r.URL.RawQuery
+       //      http.Redirect(w, r, u, http.StatusSeeOther)
+       //      return
+       // }
 
        loginChallenge := q.Get("login_challenge")
        returnTo := q.Get("return_to")
@@ -139,7 +140,16 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
                        a.cookieManager.ClearStateCookie(w)
                }
        } else {
-               response, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
+               responseFlow, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
+               setCookies(w, cookies)
+
+               if r.Header.Get("Accept") != "application/json, text/plain, */*" {
+                       u, _ := url.JoinPath(a.baseURL, "/ui/login")
+                       u = u + "?flow=" + responseFlow.Id
+                       http.Redirect(w, r, u, http.StatusSeeOther)
+                       return
+               }
+               response = responseFlow
        }
 
        if err != nil {
```

This behavior looks cleaner and does one request less. I think we should investigate and consider using this pattern across the login UI (if it actually works on all cases)

pkg/ui/handlers.go
```diff
@@ -68,15 +67,8 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
...
-	b, _ := url.Parse(baseURL)
-	k, _ := url.Parse(kratosPublicURL)
-	additionalScriptURL := ""
-	if k != nil && b != nil && k.Host != b.Host {
-		// Allowlist the kratos URL to allow the browser needs to fetch the webauthn.js script
-		additionalScriptURL = kratosPublicURL
-	}
-	return fmt.Sprintf("default-src 'self' data: https://assets.ubuntu.com; script-src 'self' %v; style-src 'self'", additionalScriptURL)
+func (a *API) getCSP(baseURL string) string {
+	return fmt.Sprintf("default-src 'self' data: https://assets.ubuntu.com; script-src 'self' %v/.well-known/webauthn.js; style-src 'self'", baseURL)
```
nsklikas, line 71: To make this work for both the juju and the docker deployments we either have to introduce a proxy in docker (probably best as it would be closer to a production env) OR make this configurable

nsklikas (commented):
Some comments, some of them are probably out of scope and should be addressed in separate PRs.

pkg/kratos/handlers.go
```diff
@@ -132,7 +133,16 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 			a.cookieManager.ClearStateCookie(w)
 		}
 	} else {
-		response, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
+		responseFlow, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
+		setCookies(w, cookies)
```
nsklikas, line 137: Cookies are set twice now, I don't remember if this is browser specific behavior, handled by golang or we can expect all browsers to simply ignore the duplicate cookies, but I think we should not do that.

pkg/kratos/handlers.go
```diff
@@ -132,7 +133,16 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
...
+
+		if r.Header.Get("Accept") != "application/json, text/plain, */*" {
```
nsklikas, line 139: nit: add explanation

```suggestion
        // If the browser was redirected here, we can't return a json object
        // so we redirect the user to the login page with the flow id appended in 
        // the query params
		if r.Header.Get("Accept") != "application/json, text/plain, */*" {
```

pkg/kratos/handlers.go
```diff
@@ -142,20 +149,29 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
...
 }
 
 func (a *API) handleCreateFlowNewSession(r *http.Request, aal string, returnTo string, loginChallenge string, refresh bool) (*client.LoginFlow, []*http.Cookie, error) {
 	// redirect user to this endpoint with the login_challenge after login
 	// see https://github.com/ory/kratos/issues/3052
+
+	cookies := r.Cookies()
+
+	// if not aal2 or both mfa and webauthn disabled clear cookies
+	if aal != "aal2" || !(a.mfaEnabled || a.oidcWebAuthnSequencingEnabled) {
```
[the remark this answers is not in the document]
nsklikas: I don't think there any exploits here as we are just not sending a session cookie so information is removed from the request. 
This also breaks adding additional FA in for oidc in case `a.mfaEnabled` and `a.oidcWebAuthnSequencingEnabled` are disabled, I understands that generally this shouldn't happen, so no problem with keeping it as it is.

pkg/kratos/handlers.go
```diff
@@ -132,7 +133,16 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 			a.cookieManager.ClearStateCookie(w)
 		}
 	} else {
-		response, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
+		responseFlow, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
```
nsklikas, line 136: add error handling

ui/components/Node.tsx
```diff
@@ -3,13 +3,13 @@ import {
   isUiNodeAnchorAttributes,
   isUiNodeImageAttributes,
   isUiNodeInputAttributes,
-  isUiNodeScriptAttributes,
+  // isUiNodeScriptAttributes,
```
nsklikas, line 6: remove this

ui/components/Node.tsx
```diff
@@ -3,13 +3,13 @@ import {
...
   isUiNodeTextAttributes,
 } from "@ory/integrations/ui";
 import { NodeAnchor } from "./NodeAnchor";
 import { NodeImage } from "./NodeImage";
 import { NodeInput } from "./NodeInput";
-import { NodeScript } from "./NodeScript";
+// import { NodeScript } from "./NodeScript";
```
nsklikas, line 12: remove this

ui/components/PageLayout.tsx
```diff
@@ -27,6 +27,7 @@ const PageLayout: FC<Props> = ({ children, title, user, isSelfServe }) => {
           href="https://assets.ubuntu.com/v1/16c27f81-COF%20favicon-16x16.png"
         />
         <title>{title}</title>
+        <script src="../.well-known/webauthn.js" />
```
nsklikas, line 30: I think this needs to be dynamic, perhaps we can fix this as part of https://github.com/canonical/identity-platform-login-ui/issues/305

pkg/kratos/handlers.go
```diff
@@ -715,14 +734,20 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// maintain previous kratos behaviour
+	setCookies(w, cookies)
+	if returnTo, ok := flow.GetReturnToOk(); ok {
```
nsklikas, line 739: same here, why does this need to be added? Is this required for webauthn?

nsklikas (dismissed):

natalian98:
I tested the changes on playground env and found some issues:
1. There is a frontend issue on adding totp as part of selfservice settings:
`Application error: a client-side exception has occurred while loading cd-iam.playground.canonical.com (see the browser console for more information).`
The POST request is completed successfully on the backend, if you reload the page you can see the totp was set.
In the console:
```
TypeError: Cannot read properties of undefined (reading 'nodes')

Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash, or a nonce ('nonce-...') is required to enable inline execution.
````

2. Similar errors can be observed for backup codes, though the nodes are not returned at all so the codes can't be set.
Password reset works fine.
3. The totp enforcement on login works but the `return_to` is lost and I get redirected to `/ui/setup_complete` instead of the app's url I tried to sign in to.

4. When adding a security key once logged in, on updating the settings flow we get the flow object instead of some redirection. I would expect to get redirected back to the settings page.
Setting webauthn works correctly on the oidc login.

**Comment by natalian98** [comment]
@edlerd  can you advise on points 1 and 2? I think they're both related to the CSP error

**Comment by edlerd** [comment]
Thanks @natalian98 for sharing the credentials for the demo environment. I was able to login and setup totp without any errors.

**Comment by edlerd** [comment]
> Steps to reproduce: Log in and go to `/ui/setup_secure`, remove and re-add the totp. I get the following errors, though I can't reproduce it every time: [Screencast from 05.09.2025 16:28:08.webm](https://github.com/user-attachments/assets/acf5e22d-aba6-4802-85b1-ee1f366447ff)

Yes, I get the same error, but the problem is not the CSP, but an unexpected backend response, that is also why the ci is failing:

image

image

image

The response shouldn't contain a redirect to, but the ui nodes for the next step in the setup.

File: pkg/kratos/handlers.go [hunk 11]
```diff
@@ -753,6 +775,30 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
+	if _, ok := body.GetActualInstance().(*client.UpdateSettingsFlowWithWebAuthnMethod); ok {
+		if state, ok := flow.State.(string); ok && state == "success" {
+			if returnTo, ok := flow.GetReturnToOk(); ok {
+				a.redirectResponse(w, r, &BrowserLocationChangeRequired{
+					RedirectTo: returnTo,
+				})
+				return
+			}
+		}
+	}
```
nsklikas, line 800: question: can you add a comment about why whit is needed? I am not sure I understand what this does, wouldn't the frontend take care of sending the user to the return_to?

pkg/kratos/handlers.go
```diff
@@ -753,6 +785,29 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
+	if flow.ReturnTo == nil {
+		if continueWith, ok := flow.GetContinueWithOk(); ok {
+			for _, c := range continueWith {
+				if continueRedirect := c.ContinueWithRedirectBrowserTo; continueRedirect != nil {
+					flow.SetReturnTo(continueRedirect.RedirectBrowserTo)
+					break
+				}
+			}
+		}
+	}
```
nsklikas: Just to be clear, are we re-writing the flow object only for the redirect in the `if` block right after? If yes, wouldn't something like this be clearer:
```
returnTo = flow.ReturnTo or flow.ContinueWithRedirectBrowserTo
redirectUserTo(returnTo)
```

This way we don't have to edit the flow object

pkg/kratos/handlers.go
```diff
@@ -753,6 +785,29 @@ func (a *API) handleUpdateSettingsFlow(w http.ResponseWriter, r *http.Request) {
...
+
+	if a.isFormPost(r) {
+		// redirect to returnTo url instead of returning a json response
+		// this maintains previous kratos behaviour on webauthn registration
+		if returnTo, ok := flow.GetReturnToOk(); ok {
```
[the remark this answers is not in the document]
nsklikas: If I understand correctly, right now we would be returning nothing (there is a `return` outside the `if` block).

Even if we were returning the flow object, the issue is that we will be returning a json to a request that expects an html. 

My gut feeling is that this should never happen, but since we wish to merge this ASAP I think that case we should simply return a server error.

pkg/kratos/handlers.go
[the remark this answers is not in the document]
nsklikas: Oh sorry I didn't see your previous response, I meant to clarify why this check is needed and if/why we need to check if sequencing and mfa are enforced.

docker-compose.dev.yml
```diff
@@ -41,6 +61,42 @@ services:
...
+      - "traefik.http.routers.kratos-admin-identities.entryPoints=web"
+      - "traefik.http.routers.kratos-admin-identities.service=kratos-admin"
+
+      - "traefik.http.routers.kratos-admin-recovery.rule=PathPrefix(`/admin/recovery`)"
+      - "traefik.http.routers.kratos-admin-recovery.entryPoints=web"
+      - "traefik.http.routers.kratos-admin-recovery.service=kratos-admin"
+
+      - "traefik.http.routers.kratos-admin-sessions.rule=PathPrefix(`/admin/sessions`)"
+      - "traefik.http.routers.kratos-admin-sessions.entryPoints=web"
+      - "traefik.http.routers.kratos-admin-sessions.service=kratos-admin"
```
nsklikas: question: why do we need to expose these?

docker/kratos/kratos.yml
nsklikas: I think we should change `serve.public.base_url` to `http://localhost/`

nsklikas (commented):
There is one final issue that we need to clarify

pkg/kratos/handlers.go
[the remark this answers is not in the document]
natalian98: What in particular was broken? I was able to set up 2fa and sign in when both flags were false.
I will remove the check unless @shipperizer you had a specific edge case in mind?

docker/kratos/kratos.yml
natalian98: It needs to include the port otherwise we get 401 responses when hitting kratos

docker-compose.dev.yml
nsklikas: In juju these rules are used by different traefik's, so I don't think it's needed here. But let's leave it as it is.

nsklikas (approved):
lgtm

PR #735, opened 2025-09-05
feat: handle duplicate identifier error in settings flow

natalian98:
This PR addresses the "Account already exists" case from https://github.com/canonical/identity-platform-login-ui/pull/731#pullrequestreview-3183585923. The backend will return an error when a duplicate account is detected.

Duplicate identifiers are detected as part of `PostSettingsPrePersistHooks`. After signing in at the external IdP, Kratos returns a 200 with an error in the settings flow `ui.nodes.messages` object (see https://www.ory.sh/docs/kratos/reference/api#tag/frontend/operation/getSettingsFlow). This is similar to how we handle some [recovery](https://github.com/canonical/identity-platform-login-ui/blob/a871b592552d7d40d6fe1a6f4ab03f6acce3375c/pkg/kratos/service.go#L373-L381) flow errors.

### DIFF
--- a/pkg/kratos/service.go
+++ b/pkg/kratos/service.go
@@ -29,6 +29,7 @@ const (
 	InvalidRecoveryCode          = 4060006
 	RecoveryCodeSent             = 1060003
 	InvalidProperty              = 4000002
+	DuplicateIdentifier          = 4000007
 	InvalidAuthCode              = 4000008
 	MissingSecurityKeySetup      = 4000015
 	BackupCodeAlreadyUsed        = 4000012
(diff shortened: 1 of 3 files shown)


pkg/kratos/service.go
```diff
@@ -325,6 +326,18 @@ func (s *Service) GetSettingsFlow(ctx context.Context, id string, cookies []*htt
...
+	// with a 4000007 error in the rendered ui messages
+	if resp.StatusCode == http.StatusOK {
+		uiMsg := flow.GetUi()
+
+		for _, message := range uiMsg.GetMessages() {
+			if message.GetId() == DuplicateIdentifier {
+				return nil, nil, fmt.Errorf("an account with the same identifier already exists, contact support")
+			}
+		}
+	}
```
BarcoMasile, line 339: note for our team in the future: I would love to see these kind of logic wrapped in a function with a proper name that lets everybody know what's needed without the necessity of having comments explaining things. Let's keep this in mind

BarcoMasile (approved):
LGTM

PR #769, opened 2025-09-30
fix: redirect to user details page when session is valid

### LINKED ISSUE #768: Repetitive login redirection even when session is valid
If you are logged in and go to `/ui/login` page, Login UI will ask to authenticate again. This happens repeatedly as we create a new login flow:

https://github.com/user-attachments/assets/354e15f8-fd3c-4186-b155-3c291031efc3

Steps to reproduce:
1. Log in
2. Go to login page again - you have to complete another login flow

imo we should instead redirect to some generic landing page when there is already a valid session that meets the aal requirements, e.g. the user management page:

https://github.com/user-attachments/assets/cb22448e-b3cf-4847-8f9c-65bef1ed528d

I will open a PR if we agree on this approach.

natalian98:
This PR changes the login flow creation logic:
- Users who are already authenticated are redirected to the User Management page instead of initiating another flow.
- Users without a session or who have not completed 2FA when required proceed with the login flow as before.

closes https://github.com/canonical/identity-platform-login-ui/issues/768

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -145,6 +145,10 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 			a.cookieManager.ClearStateCookie(w)
 		}
 	} else {
+		// aal is not always a query param, propagate it from session if aal2
+		if isSessionAAL2(session) {
+			aal = "aal2"
+		}
 		response, cookies, err = a.handleCreateFlowNewSession(r, aal, returnTo, loginChallenge, refresh)
 	}
 
@@ -219,6 +223,14 @@ func (a *API) handleCreateFlowWithSession(r *http.Request, session *client.Sessi
 	}
 
 	return response, cookies, nil
+}
+
+func isSessionAAL2(session *client.Session) bool {
+	if session == nil || session.AuthenticatorAssuranceLevel == nil {
+		return false
+	}
+
+	return *session.AuthenticatorAssuranceLevel == client.AUTHENTICATORASSURANCELEVEL_AAL2
 }
 
 func parseGenericError(err error) (*KratosErrorResponse, bool) {
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -147,11 +147,7 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 			a.cookieManager.ClearStateCookie(w)
 		}
 	} else {
-		// aal is not always a query param, for kratos flows propagate aal2 from session
-		if isSessionAAL2(session) && loginChallenge == "" {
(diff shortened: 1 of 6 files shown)


pkg/kratos/handlers.go
```diff
@@ -137,6 +144,13 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err != nil {
+		// Propagate the KratosErrorResponse so frontend can handle it
+		if kratosError, ok := parseGenericError(err); ok {
+			w.WriteHeader(http.StatusInternalServerError)
```
nsklikas, line 149: thought: perhaps a different error code should be used. It isn't a server error and it would also make the frontend logic simpler, as they can reason about the type of error based on the http status code.

nsklikas (dismissed):
lgtm

pkg/kratos/handlers.go
natalian98: Switched to 400 to follow the status returned by kratos

BarcoMasile (dismissed):

nsklikas (dismissed):

BarcoMasile (dismissed):

natalian98:
let's merge https://github.com/canonical/identity-platform-login-ui/pull/697 before this one

nsklikas (commented):
I got the following bugs:
1. If I don't complete 2fa, the flow is broken. Setup an account with 2fa and:
- Go to the log in page
- Enter username/password
- instead of completing 2fa, go back
You won't be able to log in as the filter cookie functionality is removed.

2. On first login, instead of being redirected back to the app I end up to the manage_details page.

nsklikas (dismissed):
There is a minor issue, but realistically speaking it should not occur with the current deployment config. The rest of the comments are about personal preferences, so I am approving the PR and you can choose if you wish to keep them.

nsklikas (dismissed):

pkg/kratos/handlers.go
```diff
@@ -195,12 +213,26 @@ func (a *API) handleCreateFlowNewSession(r *http.Request, aal, returnTo, loginCh
...
 	response, cookies, err := a.service.AcceptLoginRequest(r.Context(), session, loginChallenge)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to accept login request: %v", err)
+		return nil, nil, fmt.Errorf("failed to accept login request: %w", err)
 	}
 
 	return response, cookies, nil
 }
 
+func parseGenericError(err error) (*KratosErrorResponse, bool) {
```
wood-push-melon: (feel free to ignore, I might be pedantic and wrong):

We'd better follow the Go idiomatics to return `(value, error)` somehow if we can.

BarcoMasile (approved):

PR #774, opened 2025-10-06
Add security logger

nsklikas:
IAM-1699

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -48,7 +48,7 @@ func init() {
 	rootCmd.AddCommand(serveCmd)
 }
 
-func serve() int {
+func serve() error {
 
 	specs := new(config.EnvSpec)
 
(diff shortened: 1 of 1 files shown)


internal/logging/logger.go
```diff
@@ -1,60 +1,61 @@
...
-// it will need to be closed with
-// ```
-// defer logger.Desugar().Sync()
-// ```
-// to make sure all has been piped out before terminating
-func NewLogger(l string) *zap.SugaredLogger {
-	var lvl string
+type Logger struct {
+	*zap.SugaredLogger
+	security *SecurityLogger
```
shipperizer, line 13:
```suggestion
	*zap.SugaredLogger
	*SecurityLogger
```

if you were to use composition, wouldn't you be able to spare a lot of the methods defined in the Logger simply by exploiting the inheritance?

cmd/serve.go
```diff
@@ -123,13 +126,14 @@ func serve() {
...
 	// <-ctx.Done() if your application should wait for other services
 	// to finalize based on context cancellation.
-	logger.Info("Shutting down")
-	os.Exit(0)
+	return 0
+}
 
+func main() {
+	os.Exit(serve())
 }
```
shipperizer, line 139: why do we need this here?

cmd/serve.go
```diff
@@ -106,7 +108,8 @@ func serve() {
 	}
 
 	go func() {
-		if err := srv.ListenAndServe(); err != nil {
+		logger.Security().SystemStartup()
+		if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
shipperizer, line 112: is the one below caught, aka if the goroutine exits with a fatal logging, do we crash and fulfill the auditing?

nsklikas: No I think that if the server exits we don't catch it. This change is there so that we can properly catch signals, before if I killed the server (ctrl+c) the logs were not showing up.

internal/logging/logger.go
nsklikas: You mean to reuse the same logger for security and system logs? If yes, then I would like to avoid this as we (probably) may have to log security logs to a different place. Also I wanted to make it create a distinction between the 2 types of logs.

cmd/serve.go
BarcoMasile: If we want to fix that in this PR we could add a defer func calling the recover()
```
defer func() {
        if r := recover(); r != nil {
            fmt.Println("Recovered in f", r)
        }
}()
```

shipperizer (approved):

PR #828, opened 2025-12-12
Don't assume csrf token is the first node

DanielArndt:
The behaviour seems to be inconsistent in the latest version of kratos, so we adopt the `find()` pattern used elsewhere in the code.

Fixes https://github.com/canonical/kratos-operator/issues/579


BarcoMasile (commented):
Can you write one function to get the CSRF token and reuse it across the components..? So we don't have code duplication

BarcoMasile (approved):

PR #845, opened 2026-02-02
IAM-1920: add pop to AMR to support Salesforce integration

shipperizer:


nsklikas (approved):
lgtm, non-blocking: if you want remove the `oidcWebAuthnSequencingEnabled` check as it may introduce issues in the future and it is not 100% clear from reading the code why it is required.

PR #847, opened 2026-02-03
WD-33303 add countdown for resend code in verification flow

abhigyanghosh30:
Done:
- Added a timeout to the resend code button
- Also added a few error notifications

QA:
Setup the repo according to the doc: https://docs.google.com/document/d/15PrLDpaERf7FmyHwZl4GlFLPIxaX2zF2rKENxCcKgNU/edit?tab=t.0

- Visit /ui/verification
- Enter your email
- Enter a fake code
    - Should show an error
- Try to resend code
    - Should resend the code
    - Should disable the link with a 60s timer
- Visit localhost:4436 to get the correct code
- Enter the correct code
    - Should succeed

### DIFF
--- a/ui/components/CountDownText.tsx
+++ b/ui/components/CountDownText.tsx
@@ -1,0 +1,43 @@
+import React, { useState, useEffect } from "react";
+import { Notification } from "@canonical/react-components";
+
+const CountDownText = ({
+  wrapperText = "",
+  initialSeconds = 10,
+}: {
+  wrapperText?: string;
+  initialSeconds?: number;
+}) => {
+  const [seconds, setSeconds] = useState(initialSeconds);
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      setSeconds((prev) => {
+        const next = prev - 1;
+        if (next <= 0) {
+          clearInterval(timerId);
+        }
+        return next;
+      });
+    }, 1000);
+    return () => clearInterval(timerId);
+  }, [initialSeconds]);
+
+  if (seconds <= 0) {
+    return null;
+  }
+  return (
+    <Notification
+      inline
+      severity="positive"
+      borderless
+      className="u-no-margin--bottom"
+    >{`${wrapperText}${`${Math.floor(seconds / 60)
+      .toString()
+      .padStart(
+        2,
+        "0",
+      )}:${(seconds % 60).toString().padStart(2, "0")}`}s`}</Notification>
+  );
+};
+
+export default CountDownText;
--- a/ui/components/EmailVerificationPrompt.tsx
+++ b/ui/components/EmailVerificationPrompt.tsx
@@ -1,0 +1,13 @@
+import React from "react";
+
+export const EmailVerificationPrompt: React.FC<{ email: string }> = ({
+  email,
+}) => {
+  return (
+    <p className="u-text--muted">
+      An email with a verification code has been sent to {email}. Please check
+      your inbox and enter the code below. If the email is not received, ensure
+      the address is correct.
+    </p>
+  );
+};
--- a/ui/components/NodeInputSubmit.tsx
+++ b/ui/components/NodeInputSubmit.tsx
@@ -1,8 +1,9 @@
 import { getNodeLabel } from "@ory/integrations/ui";
 import { Button } from "@canonical/react-components";
 import { NodeInputProps } from "./helpers";
-import React, { Component, FC } from "react";
+import React, { FC } from "react";
 import { getProviderImage } from "../util/logos";
+import { isResendVerificationCode } from "../util/constants";
 
 export const NodeInputSubmit: FC<NodeInputProps> = ({
   node,
@@ -40,15 +41,17 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
 
   const beforeComponent = (
     node.meta.label?.context as {
-      beforeComponent: Component;
+      beforeComponent: React.ReactNode;
     }
   )?.beforeComponent;
 
   const afterComponent = (
     node.meta.label?.context as {
-      afterComponent: Component;
+      afterComponent: React.ReactNode;
     }
   )?.afterComponent;
+
+  if (isResendVerificationCode(node)) return null;
 
   return (
     <>
--- a/ui/components/NodeInputText.tsx
+++ b/ui/components/NodeInputText.tsx
@@ -1,7 +1,13 @@
-import { getNodeLabel } from "@ory/integrations/ui";
-import { Input } from "@canonical/react-components";
-import React, { FC, useEffect, useMemo } from "react";
+import { getNodeId, getNodeLabel } from "@ory/integrations/ui";
+import { Button, Input } from "@canonical/react-components";
+import React, { FC, useCallback, useEffect, useMemo } from "react";
 import { NodeInputProps } from "./helpers";
+import { FlowContext } from "../context/FlowContext";
+import {
+  isResendVerificationCode,
+  isVerificationCodeInput,
+} from "../util/constants";
+import { UiNodeInputAttributes } from "@ory/client/api";
 
 export const NodeInputText: FC<NodeInputProps> = ({
   attributes,
@@ -16,6 +22,9 @@ export const NodeInputText: FC<NodeInputProps> = ({
     attributes.value as string,
   );
 
+  const [isDirty, setIsDirty] = React.useState(false);
+
+  const flow = React.useContext(FlowContext);
   const urlParams = new URLSearchParams(window.location.search);
   const isWebauthn = urlParams.get("webauthn") === "true";
   const ucFirst = (s?: string) =>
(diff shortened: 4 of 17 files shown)


ui/components/EmailVerificationPrompt.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+export const EmailVerificationPrompt: React.FC = () => {
+  return (
+    <p className="u-text--muted">
+      An email with a verification code has been sent to . Please check your
```
BarcoMasile: issue: missing variable here

ui/components/NodeInputSubmit.tsx
```diff
@@ -50,6 +61,19 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
...
+    }
+  }, []);
+
+  useEffect(() => {
+    if (node.meta.label?.id === 1070008) {
+      return;
+    }
+    setButtonDisabled(attributes.disabled || disabled);
+  }, [attributes.disabled, disabled]);
+
```
BarcoMasile: issue(readability): it's important to make sure everybody understands what these codes are about, please add contants to the constants file and use those, this way we know what this labels are about

ui/pages/verification.tsx
```diff
@@ -115,13 +152,46 @@ const Verification: NextPage = () => {
...
 
+  const lookupFlow = useMemo(() => {
+    if (!flow) {
+      return flow;
+    }
+    return {
+      ...flow,
+      ui: {
+        ...flow.ui,
+        nodes: flow.ui.nodes.map((node) => {
```
BarcoMasile: The `map` approach here is a little overkill, since you're applying two different processing logics to the whole array with if statements to limit your action. The proper way would be to extract the nodes you're interested in, process them as you need, and then fit them back in the array. But I understand that doesn't look as "in order".

issue: I'm ok with the map approach, but the way the logic here mixes 2 different processing needs some refactor to make things clearer, for example through the use of ad-hoc functions

ui/pages/verification.tsx
```diff
@@ -93,14 +94,50 @@ const Verification: NextPage = () => {
...
+            data.ui.messages?.forEach((message) => {
+              if (message.type === "error") {
+                codeUiNode?.messages.push({
+                  id: message.id,
+                  type: "error",
+                  text: "Verification code incorrect. Check your email or resend the code.",
+                });
+              }
+            });
+          }
```
BarcoMasile: issue(readibility): same as above, please make this clearer and more readable

ui/components/EmailVerificationPrompt.tsx
abhigyanghosh30: Fixed.

ui/components/NodeInputSubmit.tsx
```diff
@@ -15,6 +15,17 @@ export const NodeInputSubmit: FC<NodeInputProps> = ({
   const isProvider = attributes.name === "provider";
   const provider = attributes.value as string;
   const image = getProviderImage(provider);
+  const [buttonDisabled, setButtonDisabled] = useState<boolean>(false);
+
+  const DISABLE_TIMEOUT = 10000; // 10 seconds
```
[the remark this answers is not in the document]
abhigyanghosh30: Moved the timeout logic to the verification component

ui/pages/verification.tsx
natalian98: Email address validation is not working properly. For example, if you type `test@e` no error is displayed.

ui/pages/verification.tsx
```diff
@@ -1,20 +1,37 @@
...
+import {
+  isResendVerificationCode,
+  isVerificationCodeInput,
+} from "../util/constants";
+import CountDownText from "../components/CountDownText";
+
+function setFlowIDQueryParam(router: NextRouter, flowId: string) {
+  void router.push(
+    {
+      pathname: router.pathname,
```
nsklikas: issue: This is wrong. AFAICT you are setting the browser path based on the pathname in the nextjs router. These 2 paths are not the same, nextjs cannot know the actual route of the page as most of the time the files are served using a reverse proxy.

For example if you go to the verification page and try to refresh you will get a 404 error.

ui/components/NodeInputText.tsx
```diff
@@ -60,30 +103,78 @@ export const NodeInputText: FC<NodeInputProps> = ({
...
+              (resendButton?.attributes as UiNodeInputAttributes).disabled
+            }
+          >
+            Resend code
+          </Button>
+        </>
+      );
+    }
+    return getNodeLabel(node);
+  }, [node, flow]);
```
nsklikas: issue: Not sure if this is the correct approach. If I go to the verification page and input my email, I will get an email and the `Resend code` button will be unclickable. BUT if I just refresh the page (I have to change the path because of the bug with `router.push`), I can click the resend button even if 1min has not passed. Not sure how this should be implemented.

nsklikas (commented):
Not a frontend engineer so I will not comment on the code, but there are some issues with the implementation.

Run some tests on my own and also used antigravity to test it, this is the report generated from antigravity:

## Summary
The PR successfully implements the core "countdown" and "disable button" logic. However, several critical issues regarding state persistence, routing, and user experience were identified during testing. These issues make the feature fragile and easily bypassable.
## ✅ Verified Features
- **Countdown Timer:** Correctly initialized when valid code is sent.
- **Resend Disable:** Button is correctly disabled during countdown.
- **Input Validation:** Invalid codes are correctly rejected with appropriate error messages.
- **Success Flow:** Valid codes lead to successful verification and redirection.
## ⚠️ Issues & Change Requests
### 1. Countdown Persistence (Critical)
- **Issue:** The countdown functionality relies entirely on the temporary page state. Reloading the page resets the state, effectively clearing the countdown.
- **Impact:** Users can bypass the resend throttle by simply refreshing the browser.
- **Recommendation:** Persist the timestamp (e.g., in `localStorage` or `sessionStorage`) so the countdown resumes correctly after a reload.
### 2. Routing & URL Structure
- **Issue:** During the verification flow, the browser URL redirects to `/verification?flow=...`, dropping the necessary `/ui` prefix.
- **Impact:** If a user reloads the page while in the flow, they receive a **404 Not Found** error because the application is served under `/ui`.
- **Recommendation:** Ensure all redirects and router pushes maintain the `/ui` prefix.
### 3. Back Button Behavior
- **Issue:** Clicking the browser's "Back" button from the code entry screen changes the URL but does not update the UI (it remains on the code entry screen). Subsequent clicks lead to 404 errors.
- **Impact:** Users cannot navigate back to correct a typo in their email address.
- **Recommendation:** Ensure the router history is correctly managed (e.g., using `replace` instead of `push` where appropriate, or handling popstate events) to allow proper navigation.
### 4. Response Message UX
- **Issue:** The success notification displays: *"You will be redirected to /ui/manage_details"*.
- **Impact:** This exposes internal URL paths to the end-user, looking unpolished.
- **Recommendation:** Update the text to a user-friendly message, such as *"You will be redirected to your dashboard"* or simply *"Redirecting..."*.
### 5. Email Link
- **Note:** The verification link sent in emails (`/self-service/verification...`) is incorrect and leads to a 404.

ui/util/flowHelper.ts
```diff
@@ -1,11 +1,3 @@
...
-export function setFlowIDQueryParam(router: NextRouter, flowId: string) {
-  void router.push(
-    {
-      query: {...router.query, flow: flowId},
-    },
-    undefined,
-    {shallow: true},
-  );
+export function setFlowIDQueryParam(flowId: string) {
+  window.history.replaceState(null, "", `?flow=${flowId}`);
```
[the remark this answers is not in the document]
BarcoMasile: Copilot is wrong because the helper function actually merges any "query" object already present with the flow query parameter, so it works as expected.

ui/components/CountDownText.tsx
```diff
@@ -0,0 +1,38 @@
...
+  if (seconds <= 0) {
+    return null;
+  }
+  return (
+    <Notification
+      inline
+      severity="positive"
+      borderless
+      className="u-no-margin--bottom"
+    >{`${wrapperText}${seconds >= 60 ? `${Math.floor(seconds / 60)}:${(seconds % 60).toString().padStart(2, "0")}` : `${seconds.toString().padStart(2, "0")}s`}`}</Notification>
```
BarcoMasile: issue: this shows time in a weird way for the 1:00 minute time remaining, in particular it shows

"00:1:00" because the "00" is hardcoded in the parent component

ui/components/NodeInputText.tsx
```diff
@@ -16,6 +22,7 @@ export const NodeInputText: FC<NodeInputProps> = ({
     attributes.value as string,
   );
 
+  const flow = React.useContext(FlowContext);
```
BarcoMasile: nitpick: as I appreacitate you following my suggesiton from our previous meeting, I would advise we follow the "react" way of doing data sharing, though not only a context provider, but also a custom hook, even as basic as it would be for this case

ui/pages/verification.tsx
```diff
@@ -27,14 +29,29 @@ function setFlowIDQueryParam(router: NextRouter, flowId: string) {
...
+  const RESEND_CODE_TIMEOUT = 10000; // 10 seconds
 
+  const [resendDisabled, setResendDisabled] = useState<boolean>(false);
+  const disableButtonWithTimeout = () => {
+    setResendDisabled(true);
+    const timer = setTimeout(() => {
+      setResendDisabled(false);
+    }, RESEND_CODE_TIMEOUT);
+    return () => clearTimeout(timer);
+  };
```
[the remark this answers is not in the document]
BarcoMasile: Can you address this? We talked about this last friday

BarcoMasile (changes requested):
## Summary
We still have a certain amount of issues that really need addressing.
Also, please address comments from the past, there are comments from two weeks ago that went unnoticed.

## UI issues
Some issues
- erroring out the email input field before on blur instead of at the typing of each character
- the time display in the countdown shows a weird "00:1:00" at the beginning of the timer, you should drop the hardcoded text from the parent component and deal with the whole "time" string only in the countdowntext component, that's going to support even different times
Here's the picture + screencast:


image


https://github.com/user-attachments/assets/d2448eeb-0a29-4901-9de9-c14af51ba081

Did you try to run the code before pushing?

ui/pages/verification.tsx
```diff
@@ -82,15 +98,77 @@ const Verification: NextPage = () => {
...
+                        wrapperText="Code sent. You can request again in 00:"
+                        key={new Date().toISOString()}
+                      />
+                    ),
+                  },
+                },
+              } as UiNodeMeta;
+            }
+            // Disable resend button for 10 seconds
+            disableButtonWithTimeout();
```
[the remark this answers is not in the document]
abhigyanghosh30: removed the return value as it is not required

PR #865, opened 2026-03-16
Enforce verification on login

natalian98:
This PR adds optional verification enforcement on login.

- Instead of relying on the kratos post-login hook `require_verified_address`, I implemented the logic in Login UI, following the pattern of MFA enforcement.
The reason is that while Kratos has a session hook for the registration -> verification flow, it currently lacks a similar hook for the login -> verification scenario. 
When the `require_verified_address` hook blocks the login, Kratos aborts the login flow and does not mint a continuation cookie. If we attempt to resume the login flow after a successful verification, Kratos returns 400 ("No session was found for this flow. Please retry the authentication") because it lost the login context. In that case we'd have to make the user go through login again, repeating the steps that were already completed before verification.
By implementing the logic in the login handlers, we can keep the session, make the user verify, and then resume the login flow from where it was interrupted.
- Added new config `VERIFICATION_ENABLED` (defaults to false). The reason for introducing it is that enabling verification is a breaking change. An SMTP server is required to enable it, otherwise existing non-verified local users would be blocked from signing in.

## Testing
Make sure to `export VERIFICATION_ENABLED="true"` and run the local docker-compose setup. Use the `test@example.com / test` identity or create another one:
```
curl -X POST http://localhost:4434/admin/identities \
  -H "Content-Type: application/json" \
  -d '{
    "schema_id": "default",
    "traits": {
      "email": "test1@example.com",
      "name": "Test1",
      "surname": "Example"
    },
    "credentials": {
      "password": {
        "config": {
          "password": "test"
        }
      }
    }
  }'
```
Then go to the login page and sign in. If the account was not previously verified, you will be redirected to the verification screen and asked to provide a code. Upon successful verification, you will need to complete MFA setup (if enabled), and finally redirected back to the oidc app.

[Screencast from 16.03.2026 12:11:54_verify_login.webm](https://github.com/user-attachments/assets/b2f68e99-7315-46a4-9be4-85c246669430)
[Screencast from 16.03.2026 12:12:48_complete_setup.webm](https://github.com/user-attachments/assets/c5d1379e-8b6d-4287-af5e-e09f19886500)

### DIFF
--- a/pkg/kratos/handlers.go
+++ b/pkg/kratos/handlers.go
@@ -21,6 +21,7 @@ import (
 	"github.com/canonical/identity-platform-login-ui/pkg/ui"
 )
 
+const VERIFICATION_REQUIRED = "verification_required"
 const TOTP_REGISTRATION_REQUIRED = "totp_registration_required"
 const WEBAUTHN_REGISTRATION_REQUIRED = "webauthn_registration_required"
 const RegenerateBackupCodesError = "regenerate_backup_codes"
@@ -30,6 +31,7 @@ const LOGIN_UI_STATE_COOKIE = "login_ui_state"
 const SECURITY_CSRF_VIOLATION_ERROR = "security_csrf_violation"
 
 type API struct {
+	verificationEnabled           bool
 	mfaEnabled                    bool
 	oidcWebAuthnSequencingEnabled bool
 	service                       ServiceInterface
@@ -105,6 +107,18 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	// TODO: We need to send a different content-type to CreateBrowserLoginFlow in order to avoid this bug.
 	session, _, _ := a.service.CheckSession(r.Context(), r.Cookies())
 	if session != nil {
+		shouldEnforceVerification, unverifiedEmail, err := a.shouldEnforceVerificationWithSession(r.Context(), session)
+		if err != nil {
+			a.logger.Errorf("failed check for verification status: %v", err)
+			http.Error(w, "Failed check for verification", http.StatusInternalServerError)
+			return
+		}
+		if shouldEnforceVerification {
+			flowCookie := FlowStateCookie{LoginChallengeHash: hash(loginChallenge)}
+			a.verificationRedirect(w, r, returnTo, flowCookie, unverifiedEmail)
+			return
+		}
+
 		shouldEnforceMfa, err = a.shouldEnforceMFAWithSession(r.Context(), session)
 
 		if err != nil {
@@ -439,6 +453,14 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 	lc := u.Query().Get("login_challenge")
 	flowCookie := FlowStateCookie{LoginChallengeHash: hash(lc)}
 
+	shouldEnforceVerification, unverifiedEmail, err := a.shouldEnforceVerification(r.Context(), cookies)
+	if err != nil {
+		err = fmt.Errorf("verification enforce check error: %v", err)
+		a.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
 	shouldEnforceMfa, err := a.shouldEnforceMFA(r.Context(), cookies)
 	if err != nil {
 		err = fmt.Errorf("enforce check error: %v", err)
@@ -457,6 +479,12 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
 	}
 
 	setCookies(w, cookies)
+
+	if shouldEnforceVerification {
+		// redirect to verification ui and pass the original returnTo so the flow continues after verification
+		a.verificationRedirect(w, r, *loginFlow.ReturnTo, flowCookie, unverifiedEmail)
+		return
+	}
 
 	if shouldRegenerateBackupCodes {
 		a.lookupSecretsSettingsRedirect(w, r, flowId, *loginFlow.ReturnTo, flowCookie)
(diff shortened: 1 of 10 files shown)


pkg/kratos/service.go
```diff
@@ -1538,6 +1538,57 @@ func (s *Service) HasNotEnoughLookupSecretsLeft(ctx context.Context, id string)
...
+		if method.Method != nil && *method.Method == "password" {
+			pwdLogin = true
+			break
+		}
+	}
+
+	// skip verification enforcement for other methods
+	if !pwdLogin {
+		return false, "", nil
+	}
```
nsklikas, line 1561: question: why don't we want to enforce verification for other methods?

nsklikas (commented):
I have a question:
~Why are we using the kratos hook? Wouldn't it be preferable if we don't re-implement it on our own?~

Nvm re-read the description

pkg/kratos/handlers.go
```diff
@@ -105,6 +107,18 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	// TODO: We need to send a different content-type to CreateBrowserLoginFlow in order to avoid this bug.
 	session, _, _ := a.service.CheckSession(r.Context(), r.Cookies())
 	if session != nil {
+		shouldEnforceVerification, unverifiedEmail, err := a.shouldEnforceVerificationWithSession(r.Context(), session)
+		if err != nil {
+			a.logger.Errorf("Failed check for verification: %v", err)
```
BarcoMasile:
```suggestion
			a.logger.Errorf("failed check for verification-enabled status: %v", err)
```
or something like that, since what's failing is the method to check if it's required
also, error texts are usually not capitalized in Go (although in this codebase we have a mix
but other text in this pr is lowercase so I'd stick with that)

pkg/kratos/handlers.go
```diff
@@ -105,6 +107,18 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
 	// TODO: We need to send a different content-type to CreateBrowserLoginFlow in order to avoid this bug.
 	session, _, _ := a.service.CheckSession(r.Context(), r.Cookies())
 	if session != nil {
+		shouldEnforceVerification, unverifiedEmail, err := a.shouldEnforceVerificationWithSession(r.Context(), session)
```
BarcoMasile, line 110: question: my cuorisity, what happens if we don't do this check in the createFlow method and keep it only in the update?

pkg/kratos/handlers.go
```diff
@@ -515,6 +543,80 @@ func (a *API) handleUpdateFlow(w http.ResponseWriter, r *http.Request) {
...
+	shouldEnforceVerification, unverifiedEmail, err := a.service.RequireVerificationForEmail(ctx, session)
+	if err != nil {
+		a.logger.Debugf("verification check failed: %v", err)
+		return false, "", err
+	}
+
+	return shouldEnforceVerification, unverifiedEmail, nil
+}
+
+func (a *API) verificationSettingsRedirect(w http.ResponseWriter, r *http.Request, returnTo string, flowStateCookie FlowStateCookie, email string) {
```
BarcoMasile: question(curiosity): why the word "settings" in the name?

nsklikas:
idea to reduce the custom code:
Instead of redirecting to the verification page at all, intercept the 403 `session_verified_address_required` in the frontend and show an inline message on the login page — e.g. "Your email address is not yet verified. Please check your inbox before signing in." with a link to resend the verification code. This avoids any redirect/re-auth loop entirely, at the cost of a slightly worse UX (the user has to verify and then come back and log in manually).

ui/pages/verification.tsx
```diff
@@ -109,12 +110,48 @@ const Verification: NextPage = () => {
...
+              setFlow(updateRes.data);
+              setFlowIDQueryParam(String(updateRes.data.id));
+              disableButtonWithTimeout();
+
+              // Clean up the URL
+              const url = new URL(window.location.href);
+              url.searchParams.delete("email");
+              void router.replace(url.pathname + url.search, undefined, {
+                shallow: true,
+              });
```
BarcoMasile: issue: you're not using the router correctly
the function replace allows to pass an object with query parameters, here's the [doc](https://nextjs.org/docs/pages/api-reference/functions/use-router#routerreplace), let's not reinvent the wheel

nsklikas:
I think there is an issue on the current implementation, try to:
1) go to login page
2) authenticate
3) get redirected to verification page
4) instead of inserting the verification code on the already open page, you click the email link (which is still broken)
5) verification is complete, but now the login tab is stuck on the verification page

Issues:
- Refreshing the login tab leads to a broken success page, you have to go back for the flow to continue
- The message in the verification page is too big. I remember that I had left a comment in the verification PR (to not show the url but instead print some text telling the user that they will be redirected), but I am not sure what was implemented in the end

pkg/kratos/service.go
natalian98: We currently only support registration via password and oidc. As agreed, we will skip verification for the oidc method, since external providers already handle verification on their end. This leaves only the password method. If the need ever arises, we can enforce it for other methods as well.

pkg/kratos/handlers.go
natalian98: If you abandon the tab where you started the login flow and open a new one, a new login flow is created. Since there is already a session (even with aal1), `handleCreateFlow`  would move on to the totp/webauthn enforcement checks and omit verification. So we need to enforce it on both create and update login

BarcoMasile: Thanks for the clarification

pkg/kratos/handlers.go
natalian98: sorry, my bad. Fixed

shipperizer (dismissed):
as discussed, let's open an issue to. Implement this via Kratos hooks

BarcoMasile (approved):

natalian98:
> Refreshing the login tab leads to a broken success page, you have to go back for the flow to continue

As decided during the daily, we will not implement the flow state polling approach and instead remove the link from email template.

> The message in the verification page is too big. I remember that I had left a comment in the verification PR (to not show the url but instead print some text telling the user that they will be redirected), but I am not sure what was implemented in the end

Fixed, only the url without parameters will be displayed

nsklikas (approved):

PR #866, opened 2026-03-16
Wd 33307 implement the registration flow according to design

abhigyanghosh30:
Done:
- Email and password input pages
- Redirect from registration to verification flow correctly

TBD:
- Redirect from verification to setup passkey


ui/pages/verification.tsx
```diff
@@ -181,7 +181,7 @@ const Verification: NextPage = () => {
             } else {
               const timer = setTimeout(() => {
                 clearTimeout(timer);
-                window.location.href = "/ui/manage_details";
+                void router.push("/secure_account");
```
BarcoMasile: Question: have you tried this without the `basePath` config in the next.config.js file?

ui/components/NodeInputEmail.tsx
```diff
@@ -20,9 +20,11 @@ export const NodeInputEmail: FC<NodeInputProps> = ({
...
     : message;
 
-  if (value == undefined) {
-    void setValue(defaultValue);
-  }
+  useEffect(() => {
+    if (value == undefined) {
+      void setValue(defaultValue);
+    }
+  }, []);
```
BarcoMasile, line 27: question: how come you want this to run only once?

ui/components/Field.tsx
```diff
@@ -0,0 +1,228 @@
...
+}: Props): React.JSX.Element => {
+  const labelNode = generateLabel(
+    forId,
+    required,
+    label,
+    labelClassName,
+    stacked,
+  );
+
+  const content = generateContent({
```
BarcoMasile, line 198: issue: adopt `useMemo`

ui/components/Field.tsx
```diff
@@ -0,0 +1,228 @@
...
+  label,
+  labelClassName,
+  labelFirst = true,
+  required,
+  stacked,
+  success,
+  validationId,
+  ...props
+}: Props): React.JSX.Element => {
+  const labelNode = generateLabel(
```
BarcoMasile, line 190: issue: adopt `useMemo`

ui/components/Field.tsx
```diff
@@ -0,0 +1,228 @@
...
+  labelNode,
+  help,
+  helpClassName,
+  error,
+  caution,
+  success,
+  validationId,
+  helpId,
+  isTickElement,
+}: Partial<Props> & {
```
BarcoMasile: Question: isn't `Partial<>` redundant here? All Props fields are already all optional

ui/components/PasswordInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }, [takeFocus, takeFocusDelay]);
+
+  const input: ReactNode = (
+    <input
+      className={classNames("p-form-validation__input", className)}
+      ref={inputRef}
+      type={type}
+      {...commonProps}
+    />
+  );
```
BarcoMasile: issue: adopt `useMemo` please

ui/components/PasswordToggle.tsx
```diff
@@ -1,8 +1,69 @@
...
+     * Delay takeFocus in milliseconds i.e. to let animations finish
+     */
+    takeFocusDelay?: number;
+    /**
+     * Optional class(es) to pass to the wrapping Field component
+     */
+    wrapperClassName?: string;
+  },
+  InputHTMLAttributes<HTMLInputElement>
+>;
```
BarcoMasile: issue(DRY): this is not DRY, you can share with the definitions in PasswordInput.tsx and extrapolate the common parts

ui/pages/register.tsx
```diff
@@ -190,22 +195,41 @@ const Registration: NextPage = () => {
...
+      const name = (node.attributes as UiNodeInputAttributes)["name"];
+      if (!name) {
+        return true;
+      }
+
+      return !noShowFields.includes(name);
+    });
+    const reorderedNodes: UiNode[] = filteredNodes.sort(
+      (a: UiNode, b: UiNode) => {
+        return nodeOrder.indexOf(a.group) - nodeOrder.indexOf(b.group);
```
BarcoMasile, line 199: nit: you have JSONs for free, why traversing that array everytime..?

ui/pages/_disabledPages/_secure_account.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { Col, Icon, Row } from "@canonical/react-components";
+import { useRouter } from "next/router";
+
+const secureAccount: NextPage = () => {
+  const router = useRouter();
+  return (
+    <PageLayout title="Secure your account">
+      <div
+        className="p-card clickable"
+        onClick={() => {
```
BarcoMasile, line 15: adopt `useCallback` please

ui/pages/_disabledPages/_secure_account.tsx
```diff
@@ -0,0 +1,63 @@
...
+                <strong>Set up Passkey (recommended)</strong> <br />
+                <small>
+                  Verify with your FaceID, TouchID, PIN or Security key.
+                </small>
+              </p>
+            </Col>
+          </Row>
+        </div>
+      </div>
+      <div
```
BarcoMasile: adopt `useCallback`

ui/pages/verification.tsx
nsklikas: This is a dead link (`/secure_account` does not exist), try to refresh the page when you get there and you will see the problem

ui/pages/verification.tsx
```diff
@@ -292,7 +292,7 @@ const Verification: NextPage = () => {
         </Notification>
         <Spinner
           text={`You will be redirected to ${
-            returnTo ? (returnTo as string).split("?")[0] : "/ui/manage_details"
+            returnTo ? (returnTo as string).split("?")[0] : "/ui/secure_account"
```
nsklikas: `returnTo` comes directly from the URL query string with no validation at all. This means anyone can craft a link like `http://localhost/ui/verification?return_to=https://example.com` and after a legitimate user completes verification they'll be silently redirected to an external site. This can be abused for phishing. We should only allow URLs that come from kratos for redirection

ui/components/RegisterPassword.tsx
```diff
@@ -0,0 +1,124 @@
...
+              redirectTo(redirectAction.redirect_browser_to, router);
+              return;
+            }
+          }
+        })
+        .catch((error) => {
+          // Handle errors, e.g., display error messages
+          console.error("Error submitting registration flow:", error);
+          setIsSubmitting(false);
+        });
```
nsklikas: If the registration request fails — network error, expired flow, server-side validation rejection — the error is only logged to the console and the submit button is re-enabled. From the user's perspective nothing happened, they have no idea why. Every other flow page feeds errors back through `setFlow` so they show up as proper UI messages. This should do the same.

ui/components/Password.tsx
```diff
@@ -45,39 +21,50 @@ const Password: FC<Props> = ({
...
-  const handlePasswordChange = useCallback(
-    (e: React.ChangeEvent<HTMLInputElement>) => {
-      if (!hasTouched) setHasTouched(true);
-      setPassword(e.target.value);
-    },
-    [hasTouched, setPassword],
-  );
+  const localValid = hasPassBlur && !isCheckFailed && password === confirmation;
+  if (isValid !== localValid) {
+    setValid(localValid);
```
nsklikas: copilot comment:

`setValid` is being called conditionally during the render pass. Updating state during render is not allowed in React — it can cause extra renders or in some cases loop. This should be inside a `useEffect` with `[localValid]` as the dependency array.

nsklikas (commented):
Good work overall. Use copilot with the setup I have pushed to the repo to run some tests. Left a few inline comments, summary of the main things below.

**Blocking**

- `verification.tsx:179` — open redirect. `return_to` is taken from the query string with no validation. You can reproduce it by navigating to `/ui/verification?return_to=https://example.com` and completing verification — you land on `example.com`. Only relative URLs should be accepted here.
- `RegisterPassword.tsx:93` — errors from `updateRegistrationFlow` are swallowed in the catch block. The user gets no feedback when the submit fails. All the other flow pages feed errors back through `setFlow`; this should do the same.

**Should fix**

- `Password.tsx:64` — `setValid` is called during render. This should be in a `useEffect`.

**Minor (feel free to disregard)**

- `secure_account.tsx:11,35` — the clickable cards are plain `div` elements. They can't be focused with Tab or activated with the keyboard. Should be `button` elements (the `useCallback` suggestion from the other reviewer doesn't address this).
- `NodeInputEmail.tsx:40–66` — `document.getElementsByClassName("p-button--positive")[0]` to toggle the submit button. This will break if the page ever has more than one positive button. Better to wire this through React state.

ui/pages/verification.tsx
abhigyanghosh30: I was unable to reproduce this issue. I can refresh and the page still shows

BarcoMasile: @abhigyanghosh30  If you run the frontend with npm run dev you have the settings in next.config.js working.
There you'll have also the `basePath` which will influence the behavior of redirections. That could be the case why you can't reproduce the issue, try commenting the `basePath` in next.config.js and run again.

ui/components/NodeInputEmail.tsx
abhigyanghosh30: We want to run it once per render. In case the field doesn't have a value coming from the props, the default value is assigned

ui/pages/verification.tsx
abhigyanghosh30: Replaced with redirectTo function which does the redirect better

ui/components/Field.tsx
abhigyanghosh30: Not required as this is a pure function. Everytime the props change, the component will be re-rendered and the funciton will run again anyways

abhigyanghosh30: For reference this component is a copy of https://github.com/canonical/react-components/blob/main/src/components/Field/Field.tsx which also has the same patterns

ui/components/Field.tsx
abhigyanghosh30: same as above

ui/components/PasswordToggle.tsx
abhigyanghosh30: fixed

ui/components/NodeInputEmail.tsx
BarcoMasile: It's clear now, thank you!

ui/components/Field.tsx
BarcoMasile: If you used `useMemo` you would memoize the result of the function, which would avoid re-rendering in case the inputs don't change. Memoizing is about avoiding the change of the reference to the variable instantiated so the re-render doesn't unnecessarily happens.

About the example you provided, they should use `useMemo` as well.

That said, I don't consider this a blocking issue so it's fine to leave it as is for now.

ui/components/Field.tsx
BarcoMasile: link to my other response to the same type of comment: https://github.com/canonical/identity-platform-login-ui/pull/866/changes#r3020325582

ui/components/RegisterPassword.tsx
abhigyanghosh30: updated the error to update the flow. Although I see there is no error reporting in case of non 400 errors.

ui/pages/verification.tsx
abhigyanghosh30: I see that the createVerificationFlow function requires a returnTo parameter. How should we handle that? I just removed it for now and it seems to work

ui/pages/verification.tsx
abhigyanghosh30: replaced the hardcoded basepaths with the basepath value from the router object which should be able to handle the redirects correctly. This was if we change it to something other than `/ui` in the nextconfig, it will still do the routing correctly

ui/components/Password.tsx
abhigyanghosh30: updated

ui/pages/verification.tsx
nsklikas: Don't remove the `returnTo` — we need it for the OIDC case.

The fix is to keep passing it to `createBrowserVerificationFlow` (line 115 already does this correctly). Kratos validates the value against its allowed redirect URLs and stores the validated URL on the flow object. The bug is one step later: on `passed_challenge` the redirect uses the raw query param `returnTo` instead of reading `flow.return_to`. That is what causes the open redirect — the `return_to` URL is never going through Kratos. The fix is to use `flow.return_to` instead, the same way `setup_complete.tsx` does it.

This covers both cases naturally:
1. local flow (user visiting their profile page, link from email): `flow.return_to` is not set → fall back to the default page
2. OIDC-initiated flow: the registration was started with a `return_to` pointing to the third-party app → Kratos threads it through to the verification flow → `flow.return_to` resolves to the original app

It is also fine to keep the default `"/ui/manage_details"` fallback.

One thing worth checking (probably out of scope for now): does Kratos propagate `return_to` from the registration flow onto the verification flow it creates internally (the one we get via `continue_with.show_verification_ui`)? If it does, we don't need to do anything else. If it doesn't, we'd need to pass the registration flow's `return_to` explicitly when redirecting to verification.

ui/pages/verification.tsx
BarcoMasile: > replaced the hardcoded basepaths with the basepath value from the router object which should be able to handle the redirects correctly

can you make sure it actually does? so we remove doubt @abhigyanghosh30

shipperizer:
@abhigyanghosh30 please rebase on main, clean the merge commits and potentially rewrite the history in a more concise way as those messages end up in our changelog

abhigyanghosh30:
@BarcoMasile @shipperizer @nsklikas @natalian98 Closing this one and opening a new PR https://github.com/canonical/identity-platform-login-ui/pull/872 which has the rebasing and cherrypicking done correctly

PR #869, opened 2026-04-01
Multitenancy login flow

nsklikas:
### Summary

Adds multi-tenancy support to the login flow. When `MULTI_TENANCY_ENABLED=true` and a `TENANTS_SERVICE_URL` is configured, the login UI resolves which tenants a user belongs to immediately after the identifier-first step and routes them accordingly.

### How it works

After the user enters their email and the identifier-first step advances, the frontend calls `POST /api/v0/tenants/resolve`. The backend fetches the Kratos flow to extract the email, queries the Tenant Service, and returns one of three outcomes:

| Tenants | Outcome |
|---------|---------|
| 0 | Empty `{}` — frontend proceeds to credentials directly |
| 1 | Backend auto-stores the tenant in the encrypted cookie and returns `{redirect_to: "/ui/login?flow=..."}` — no user interaction needed |
| 2+ | `{redirect_to: "/ui/select_tenant?..."}` — user picks a tenant, which calls `POST /api/v0/auth/tenant` to persist the selection |

The selected `tenant_id` is injected into the Kratos `transient_payload` at credential submit time, then carried through the Hydra consent step into the token hook for final validation.

### Changes

`refactor(internal/cookies)` — `AuthCookieManager` and `Encrypt` are moved out of `pkg/kratos` into a new `internal/cookies` package to avoid a circular import from `pkg/tenants`.

`feat(tenants)`:

- New `pkg/tenants` package: `GET /api/v0/tenants`, `POST /api/v0/tenants/resolve`, `POST /api/v0/auth/tenant`
- `TenantResolverInterface` in `pkg/kratos` — lets the login handler read the selected tenant without a hard dependency on the cookie implementation
- `pkg/extra`: injects `_tenant_id` into the Hydra consent accept session so the Tenant Service token hook can validate and burn the public `tenant_id` claim
- Config: `MULTI_TENANCY_ENABLED` (default `false`), `TENANTS_SERVICE_URL` (required when enabled)
- Frontend: `select_tenant.tsx`, `tenants.ts`, `tenantResolve.ts`; `login.tsx` calls resolve using `flow.oauth2_login_challenge` as a fallback for when the URL param is stripped after the identifier-first redirect

### Testing

To try this you need to build the tenant service from [canonical/tenant-service#53](https://github.com/canonical/tenant-service/pull/53)


nsklikas:
I ran `gofmt` on the repo and it introduced a bunch of changes to spaces (apparently the go standard is to use tabs instead of spaces). I can undo them and push them to a different PR if needed

nsklikas:
@copilot resolve your comments that are no longer valid

pkg/tenants/handlers.go
```diff
@@ -0,0 +1,224 @@
...
+	if body.TenantID == "" {
+		http.Error(w, "tenant_id is required", http.StatusBadRequest)
+		return
+	}
+
+	if err := a.storer.StoreTenant(w, r, body.TenantID, body.LoginChallenge); err != nil {
+		a.logger.Errorf("failed to persist tenant selection state: %v", err)
+		http.Error(w, "failed to persist tenant selection", http.StatusInternalServerError)
+		return
+	}
```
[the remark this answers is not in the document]
BarcoMasile: Any chance we do something about this in a future PR?
Even like adding the allowed tenants for the user at flow creation time and sticking them in the flow context?

ui/pages/select_tenant.tsx
```diff
@@ -0,0 +1,108 @@
...
+      .catch((err) => {
+        console.error(err);
+        setError(
+          "Failed to fetch tenants. Ensure the backend is running and the user exists.",
+        );
+      })
+      .finally(() => setLoading(false));
+  }, [router.isReady, flowId]);
+
+  const handleSelect = (tenantId: string) => {
```
BarcoMasile:
```suggestion
  const handleSelect = useCallback((tenantId: string) => {
  ...
  }, [router.query, login_challenge, tenantId, flow, fetch])
```

BarcoMasile (dismissed):
LGTM!!!

ui/pages/select_tenant.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [error, setError] = useState<string | null>(null);
+  const [loading, setLoading] = useState(false);
+
+  const submitTenantSelection = useCallback(
+    (tenantId: string) => {
+      const { flow, login_challenge } = router.query;
+
+      if (!login_challenge || typeof login_challenge !== "string") {
+        return;
+      }
```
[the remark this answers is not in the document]
BarcoMasile: this seems relevant, could we `setError` here?

pkg/tenants/service.go
```diff
@@ -0,0 +1,126 @@
...
+
+// emailFromFlow extracts the email from a login flow's "identifier" UI node.
+func emailFromFlow(flow *kClient.LoginFlow) (string, error) {
+	for _, node := range flow.Ui.Nodes {
+		if node.Type == "input" &&
+			node.Attributes.UiNodeInputAttributes != nil &&
+			node.Attributes.UiNodeInputAttributes.Name == "identifier" {
+			val, ok := node.Attributes.UiNodeInputAttributes.Value.(string)
+			if !ok || val == "" {
+				return "", fmt.Errorf("identifier value is not a string in flow %s", flow.Id)
```
BarcoMasile:
```suggestion
				return "", fmt.Errorf("identifier value is not a string or is empty in flow %s", flow.Id)
```

BarcoMasile (dismissed):

pkg/tenants/handlers.go
```diff
@@ -0,0 +1,230 @@
...
+
+// handleLookupTenants accepts an optional ?flow= query parameter. When flow is
+// provided it fetches the Kratos flow to extract the email; otherwise the
+// identity ID is read from the active Kratos session. Email is never accepted
+// as a URL parameter to prevent unauthenticated tenant enumeration.
+func (a *API) handleLookupTenants(w http.ResponseWriter, r *http.Request) {
+	flowID := r.URL.Query().Get("flow")
+
+	var (
+		tenants []Tenant
```
BarcoMasile: suggestion: switch to `[]*Tenant`

pkg/tenants/handlers.go
BarcoMasile, line 1: suggestion: I'd suggest to use some white space to separate logic portions of code to make it more readable

BarcoMasile (dismissed):

pkg/extra/service.go
```diff
@@ -45,10 +45,19 @@ func (s *Service) GetConsent(ctx context.Context, challenge string) (*hClient.OA
...
 	session := hClient.NewAcceptOAuth2ConsentRequestSession()
 	session.SetIdToken(misc.GetUserClaims(identity, *consent))
 
+	if tenantID != "" {
+		// Embed the tenant ID into the access token session under "_tenant_id".
+		// The leading underscore signals that this is an internal field: it is
+		// deliberately absent from allowed_top_level_claims in the Hydra config
+		// so it is never exposed in the issued token. The hook reads "_tenant_id"
+		// and maps it to the public "tenant_id" claim.
+		session.SetAccessToken(map[string]interface{}{"_tenant_id": tenantID})
```
shipperizer, line 61: just to understand, so `tenant_id` won't be a custom claim available in the ID token either?we kinda need that i'm afraid

pkg/tenants/handlers.go
nsklikas: fixed

ui/pages/select_tenant.tsx
```diff
@@ -0,0 +1,117 @@
...
+      typeof flowId === "string"
+        ? fetchTenantsByFlow(flowId)
+        : fetchTenantsBySession();
+    loader
+      .then((result) => {
+        setTenants(result);
+        if (result.length <= 1) {
+          submitTenantSelection(result.length === 1 ? result[0].id : "");
+        }
+      })
```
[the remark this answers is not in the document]
nsklikas: This is intentional, this is just a reference frontend implementation

pkg/kratos/handlers.go
```diff
@@ -387,23 +445,42 @@ func (a *API) handleUpdateIdentifierFirstFlow(w http.ResponseWriter, r *http.Req
...
 	}
 
-	setCookies(w, cookies)
+	setCookies(w, httpCookies)
+
+	// After the identifier-first step succeeds, check whether tenant
+	// selection is required BEFORE the user enters their password (1FA).
+	// This only applies to Hydra-initiated flows with multi-tenancy enabled.
+	if redirectTo != nil && a.tenantMgr.Enabled() {
+		loginFlow, _, flowErr := a.service.GetLoginFlow(r.Context(), flowId, r.Cookies())
```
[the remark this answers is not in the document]
nsklikas: The kratos response does not return all the cookies, this would break the code

pkg/tenants/handlers.go
```diff
@@ -0,0 +1,237 @@
...
+			return
+		}
+
+		identityID := identityIDFromSession(session)
+		if identityID == "" {
+			http.Error(w, "could not determine identity from session", http.StatusUnauthorized)
+			return
+		}
+		tenants, err = a.service.LookupTenantsByIdentityID(r.Context(), identityID)
+	}
```
shipperizer:
```suggestion
	if flowID == "" {
		var session *kClient.Session
		session, _, err = a.sessionChecker.CheckSession(r.Context(), r.Cookies())
		if err != nil {
			http.Error(w, "unauthorized", http.StatusUnauthorized)
			return
		}

		identityID := identityIDFromSession(session)
		if identityID == "" {
			http.Error(w, "could not determine identity from session", http.StatusUnauthorized)
			return
		}
		tenants, err = a.service.LookupTenantsByIdentityID(r.Context(), identityID)
	} else {
	   tenants, err = a.service.LookupTenantsByFlow(r.Context(), flowID, r.Cookies())
	}
```

pkg/tenants/handlers.go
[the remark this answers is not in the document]
shipperizer: wouldn't that expose too many info (albeit in the encrypted cookie)?
feels like https://github.com/canonical/identity-platform-login-ui/pull/869/changes#diff-34a75bc3778fd77d050dd64f246b0730f611c07a94166d3eae6ae94878b0c0bcR113-R119 should be enough

pkg/tenants/service.go
```diff
@@ -0,0 +1,159 @@
...
+	CreatedAt string `json:"created_at"`
+	Enabled   bool   `json:"enabled"`
+}
+
+// Service fetches tenant data from the external tenants API.
+// TODO: Replace the ad-hoc HTTP client with the autogenerated SDK when the
+// tenant service exposes a gRPC endpoint.
+type Service struct {
+	tenantsAPIURL string
+	httpClient    *http.Client
```
shipperizer, line 34: let's add an issue and make this a gRPC service

pkg/extra/service.go
nsklikas: It will be injected in both of them: https://github.com/canonical/hook-service/pull/225/changes#diff-9833bae118c17e3436774c0aceb9f07a9dfc7f0c78a4213c199cd2258cfb195dR169

This is just a place to store it temporarily for the hook-service to fetch it. This(`_tenant_id`) should not be exposed to the user.

pkg/tenants/handlers.go
nsklikas: fixed

pkg/tenants/login_flow.go
BarcoMasile, line 1: nitpick: seems more like a `util.go` file for utilities

ui/pages/select_tenant.tsx
```diff
@@ -0,0 +1,126 @@
...
+        body: JSON.stringify(body),
+      })
+        .then((r) => {
+          if (!r.ok) {
+            throw new Error(`failed to store tenant selection: ${r.status}`);
+          }
+          return r.json() as Promise<{ redirect_to: string }>;
+        })
+        .then((r) => {
+          window.location.href = r.redirect_to;
```
BarcoMasile, line 50: watch out, this will miss any query params from the current url, if there's something that needs to be carried
I think it is fine like this, just calling it out for reasoning

BarcoMasile (dismissed):

pkg/kratos/handlers.go
```diff
@@ -107,68 +108,112 @@ func (a *API) handleCreateFlow(w http.ResponseWriter, r *http.Request) {
...
+	// When DeferMFAChecks is set the user has a session from a previous
+	// flow (cookie doesn't match this challenge). We must still consult
+	// MustReAuthenticate because Hydra may demand re-auth (max_age=0).
+	// Only if Hydra says skip=true do we honour SelectTenant/AcceptLogin.
+	if intercept.DeferMFAChecks || (!intercept.AcceptLogin && !intercept.SelectTenant) {
+		var forceLogin bool
+		forceLogin, err = a.service.MustReAuthenticate(r.Context(), loginChallenge, session, c)
+		if err != nil {
+			a.logger.Errorf("Failed to fetch hydra flow: %v", err)
+			http.Error(w, "Failed to fetch hydra flow", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
nsklikas: This is just a formatting change, the inconsistency was already there and it's out of scope for this PR to fix it

pkg/tenants/login_flow.go
nsklikas: I would argue that the function in there is not generic enough to be called a util, maybe `login_payload.go` would be better. I'll keep it as it is unless you have a strong objection.

ui/pages/select_tenant.tsx
nsklikas: Good catch. TBH my initial assumption when creating this page was that it would be a poc and the frontend team would refactor it, but now this seems unlikely to happen.

I looked a little on how we handle this in other places and in some places we simply pass whatever the backend returned, whereas in others we carry over the query params. We may have to refactor this in the future, but I don't think any change is needed for now.

BarcoMasile (approved):

PR #882, opened 2026-04-28
fix: add flow error handler in register password view

### LINKED ISSUE #860: Late validation for existing account identifiers, no "account already exists" error displayed in UI
When attempting to create a new account with an identifier already assigned to another identity, on the second step of registration backend returns an error "an account with the same identifier already exists, contact support":

https://github.com/user-attachments/assets/36a2a0b0-1135-44d6-9c4a-66ac138ff7c6

Perhaps we should verify that the identity exists on the first registration step, before making the user enter any credentials.

abhigyanghosh30:
Fixes errors not showing up in RegisterPassword page in Registration Flow

QA Steps:
Visit /ui/register
Input test@example.com as email and press next
Enter any valid password and enter the same one in the field below
Press submit
An error should show up saying that the email is already registered
Fixes https://github.com/canonical/identity-platform-login-ui/issues/860

### DIFF
--- a/ui/components/NodeInputEmail.tsx
+++ b/ui/components/NodeInputEmail.tsx
@@ -27,7 +27,7 @@ export const NodeInputEmail: FC<NodeInputProps> = ({
   }, []);
 
   useEffect(() => {
-    setError(upstreamError);
+    setError(getError());
   }, [upstreamError]);
 
   const getError = useCallback(() => {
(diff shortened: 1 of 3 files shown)


BarcoMasile (approved):

BarcoMasile:
Just rebased on main and we can proceed to merge

PR #889, opened 2026-05-04
fix: avoid query param passing (with flow param) when switching flows

### LINKED ISSUE #874: Incorrect call to kratos after registration
After you register a new account and visit the manage details page for the first time, the "email" field is empty:

Image

The reason is that frontend makes a call to kratos instead of login ui (https://iam.blue.canonical.com/self-service/settings/flows?id=...) that returns 404.
If you go to https://iam.blue.canonical.com/ui/manage_details instead, the email is returned correctly.

BarcoMasile:
## Description
Fixing a behavior of the redirection logic used in the verification flow where the flow id for verification gets re-used for other flows causing issues in the page you get redirected to.

Fixes #874

### DIFF
--- a/ui/pages/verification.tsx
+++ b/ui/pages/verification.tsx
@@ -177,12 +177,12 @@ const Verification: NextPage = () => {
         .then(({ data }) => {
           if (data.state === "passed_challenge") {
             if (flow?.return_to) {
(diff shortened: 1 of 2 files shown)


nsklikas (approved):

natalian98 (approved):
lgtm

PR #895, opened 2026-05-21
Don't filter out Back buttons + restart flow when clicked

BarcoMasile:
## Description
Don't filter out Back buttons anymore.
When clicked, the new back button will restart the current flow by removing the `flow` query parameter from the url, representing the ID of the flow object.
Flow objects will expire and be handled by Kratos, so no issues with just creating a new one to allow users to go back (although this could be abused by a malevolent user to bother the kratos instance, but it should be negligeable).


ui/pages/reset_email.tsx
```diff
@@ -108,7 +108,7 @@ const ResetEmail: NextPage = () => {
                   text: "Reset password",
                   context: {
                     ...node.meta.label.context,
-                    beforeComponent: <BackButton />,
+                    beforeComponent: <ResetEmailBackButton />,
                   },
```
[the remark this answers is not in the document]
BarcoMasile: This is intended for that specific flow in order not to break something that is working :D at least not for this first PR.

ui/components/Node.tsx
```diff
@@ -32,6 +42,13 @@ export const Node: FC<Omit<NodeInputProps, "attributes">> = ({
     return <NodeAnchor node={node} attributes={node.attributes} />;
   }
 
+  if (isUiNodeBackButton(node.meta)) {
+    const attrs = node.attributes as UiNodeInputAttributes;
+    return (
+      <FlowBackButton disabled={attrs.disabled} text={node.meta.label?.text} />
+    );
+  }
```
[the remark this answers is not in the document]
BarcoMasile: the flow will be updated by the parent component when url is loaded without the `flow` query param.

ui/components/NavigationButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+ * Intended for use inside Kratos node metadata (reset_email page) where the
+ * component is injected into `meta.label.context.beforeComponent`.
+ */
+export const ResetEmailBackButton: FC<ResetEmailBackButtonProps> = ({
+  tabIndex,
+}) => {
+  const onClick = useCallback(() => window.history.back(), []);
+
+  return (
+    <Button tabIndex={tabIndex ?? 3} type="button" onClick={onClick}>
```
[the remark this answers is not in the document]
BarcoMasile: Back buttons haven't had a tabIndex till now (apart from the one in reset_email.tsx) so all is good for now, based on the behavior we'll see we can easily come back and provide tab indices.

ui/components/RegisterPassword.tsx
```diff
@@ -134,6 +134,7 @@ export const RegisterPassword = ({ flow, setFlow }: RegisterPasswordProps) => {
             <p className="p-form-validation__message">{flowError}</p>
           </div>
         )}
+        <FlowBackButton />
```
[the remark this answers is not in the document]
BarcoMasile: The redirection happens with the flow removed so it will not be present in the URL when the component rerenders

ui/components/Node.tsx
```diff
@@ -11,6 +11,16 @@ import { NodeInput } from "./NodeInput";
...
+import { FlowBackButton } from "./NavigationButtons";
+import { UiNodeInputAttributes, UiNodeMeta } from "@ory/client";
+
+function isUiNodeBackButton(meta: UiNodeMeta) {
+  return (
+    meta.label?.id === 1040008 &&
+    meta.label?.type === "info" &&
+    meta.label?.text === "Back"
+  );
+}
```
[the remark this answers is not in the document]
BarcoMasile: Done in next revision. Thanks

natalian98:
I found some issues, they are reproducible on blue (https://iam.blue.canonical.com/blue-cos-grafana):
1. Go to login page, click on register
2. Note that registration flow is created with correct url (including `/ui`): https://iam.blue.canonical.com/ui/register?flow=... in this step. Insert email and go to the next step.
3.  In the password step, click on the Back button
4. When going back the /ui prefix is not present, so the url becomes https://iam.blue.canonical.com/register?flow=bd7a9ed4-1c53-4b4f-9ec3-59820f0060b0 and we get 404s when getting some resources (like the logo), see:

image

(the "Sign in" link is broken as well in that case - `/login` instead of `/ui/login`)

5. Then if you attempt to register an account, after choosing a password, the Next button is not responding and the registration never completes:
image

BarcoMasile:
I have a clear idea of the issue now, it's all related to the missing basePath in our prod settings
Will have a look soon @natalian98

ui/util/constants.ts
```diff
@@ -89,3 +90,11 @@ export const isRegisterPasswordInput = (node: UiNode): node is NodeWithLabel =>
   node.meta.label?.id === ORY_LABEL_REGISTER_PASSWORD_INPUT;
 
 export const ORY_ERR_ACCOUNT_NOT_FOUND_OR_NO_LOGIN_METHOD = 4000037;
+
+export function isUiNodeBackButton(meta: UiNodeMeta) {
```
nsklikas, line 93: question: Should this also catch `1060007` (recovery back button)?

ui/components/NavigationButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export const FlowBackButton = <T,>({
+  setFlow,
+  ...props
+}: FlowBackButtonProps<T>) => {
+  const router = useRouter();
+
+  const onBack = useCallback(async () => {
+    const { flow: _flow, ...noFlowQueryParams } = router.query;
+    void _flow; // ignoring the flow id to "restart" the flow
```
nsklikas: nit: couldn't this be re-written to
```suggestion
    const { flow: _, ...noFlowQueryParams } = router.query;
```
?

ui/components/Node.tsx
```diff
@@ -32,6 +34,13 @@ export const Node: FC<Omit<NodeInputProps, "attributes">> = ({
     return <NodeAnchor node={node} attributes={node.attributes} />;
   }
 
+  if (isUiNodeBackButton(node.meta)) {
+    const attrs = node.attributes as UiNodeInputAttributes;
+    return (
+      <FlowBackButton disabled={attrs.disabled} text={node.meta.label?.text} />
```
nsklikas, line 40: question: don't we need to pass a `SetFlow` function so that it can clear the flow state?

ui/components/NavigationButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    await router.replace({ query: noFlowQueryParams });
+    if (setFlow) setFlow(undefined);
+  }, [router, setFlow]);
+
+  return (
+    <Button
+      tabIndex={props.tabIndex}
```
nsklikas, line 61: question: AFAICT this is undefined where as in `ResetEmailBackButton` `tabIndex=3`. Is this intentional? Shouldn't both be undefined?

ui/util/constants.ts
nsklikas: Or maybe it should be called `isRegistrationUiNodeBackButton`?

ui/components/NavigationButtons.tsx
```diff
@@ -0,0 +1,63 @@
...
+  ...props
+}: FlowBackButtonProps<T>) => {
+  const router = useRouter();
+
+  const onBack = useCallback(async () => {
+    const { flow: _flow, ...noFlowQueryParams } = router.query;
+    void _flow; // ignoring the flow id to "restart" the flow
+
+    await router.replace({ query: noFlowQueryParams });
+    if (setFlow) setFlow(undefined);
```
nsklikas: question (ai comment): should this be wrapped in a try/catch block in case the `replace` fails?

BarcoMasile:
> I have a clear idea of the issue now, it's all related to the missing basePath in our prod settings Will have a look soon @natalian98

Actually, I already (thought I) provided code to avoid this.
I'm investigating

nsklikas (commented):
overall lgtm, mostly nits

ui/components/NavigationButtons.tsx
BarcoMasile: It's intentional, the existing back button (used in only one page) had the tabIndex set to 3 so I made sure it wasn't changed by the refactor.

ui/components/Node.tsx
BarcoMasile: It would be nice although it works the same.
I should have put in the description some of the notes from the investigation
Basically I avoided passing the setFlow because the change would have impacted a few other things and I wanted to target the issue without changing too much

ui/util/constants.ts
BarcoMasile: Nice catch thanks, I "generalized" that check so it catches all "Back" info buttons

ui/components/NavigationButtons.tsx
BarcoMasile: Hopefully this never fails but we never know so I added it, thanks

natalian98:
The path issue is solved, but there is still an issue with registration flow. When you submit the password, we get a flow object from kratos, but the registration is not completed:
[Screencast from 2026-06-01 11-46-30.webm](https://github.com/user-attachments/assets/6c1fd05a-6890-495d-bbc4-e2b1787df4f7)

Edit: this issue only occurs when the registration hook is present. We will track it independently as it is out of scope for this PR: https://github.com/canonical/identity-platform-login-ui/issues/901

natalian98 (approved):

PR #936, opened 2026-08-21
Upgrade hydra client to v26 and remove custom device flow client

nsklikas:


docker/traefik/login-ui-routes.yml
```diff
@@ -16,12 +16,12 @@ http:
...
 
-    # /api/device
-    login-ui-public-api-router-api-device:
+    # /api/hydra
+    login-ui-public-api-router-api-hydra:
       entryPoints:
         - web
         - websecure
-      rule: "Path(`/api/device`)"
+      rule: "PathPrefix(`/api/hydra`)"
```
natalian98, line 24: nit: we probably want to update the [skill](https://github.com/canonical/identity-platform-login-ui/blob/c0f0e24ce26435e31fb7c45667ed2e82b86347de/.github/skills/debug-browser-errors/SKILL.md?plain=1#L98) as well, it's good to keep it up to date if we switch back to copilot

natalian98 (dismissed):
lgtm, let's also remember to update the login ui [routes](https://github.com/canonical/identity-platform-login-ui-operator/blob/36b546b21dcbc0ae14ae3154acaf257a497deead/templates/public-route.json.j2#L36-L54) (both on main and on track/istio branch)

natalian98 (approved):

PR #970, opened 2026-09-09
Separation of the frontend part of the project

BarcoMasile:
## Description

This pull request completes the separation and complete removal of the Next.js frontend user interface from this backend repository. The frontend now lives and is maintained in its own dedicated repository.

With these changes:
- **Frontend Code Removal**: Deleted the entire `ui/` directory containing Next.js source code, components, pages, Nginx configurations, UI Dockerfile, and Playwright E2E browser tests.
- **Backend Architecture**: The Go server now acts strictly as a stateless REST API / BFF microservice without any embedded static assets (`//go:embed ui/dist`), static file handlers, or Node.js build steps.
- **Build & CI Simplification**: Removed `npm-build`, `test-e2e`, and `test-e2e-debug` targets from the `Makefile`, removed the `frontend-ui` service from `docker-compose.yml`, updated `.dockerignore` and `renovate.json` to drop NPM package rules, and updated CI workflows (`ci.yaml`, `tiobe.yaml`) to run purely Go testing and compilation.
- **Documentation**: Updated developer instructions, `README.md`, `CONTRIBUTING.md`, and Copilot prompts to reflect the backend-only project structure and workflows (`make build`).

---

### 1. Frontend Code & Asset Clean-up
- Deleted `ui/` directory and obsolete instructions (`.github/instructions/frontend.instructions.md`, `playwright.instructions.md`).
- Deleted Playwright E2E workflow (`.github/workflows/e2etest.yaml`). E2E browser tests are now located in the standalone frontend repository.

### 2. Build Tooling & Orchestration
- Removed `npm-build`, `test-e2e`, and `test-e2e-debug` targets from `Makefile`.
- Removed `frontend-ui` container service from `docker-compose.yml`.
- Cleaned up `ui/` and `node_modules` entries from `.dockerignore`.
- Removed NPM / UI package manager rules from `renovate.json`.
- Removed Node.js setup and `npm-build` steps from `.github/workflows/tiobe.yaml` and removed `e2e-test` job dependency from `.github/workflows/ci.yaml`.

### 3. Documentation & Developer Tooling
- Updated `README.md` and `CONTRIBUTING.md` build instructions (`make build`).
- Updated `.github/copilot-instructions.md` and `.github/prompts/setup-dev-env.prompt.md` for backend-only setup.

---

## How Has This Been Tested?

- **Go Unit Tests & Vet**: Ran `make test` — all Go unit tests passed cleanly (`ok`) and `go vet` reported zero issues.
- **Go Compilation**: Ran `make build` / `go build -o app ./` — compiled the standalone Go binary `./app` successfully without errors or static asset dependencies.


ui/Dockerfile
nsklikas, line 1: idea (not sure if possible): wouldn't it be simpler to just compile the frontend and mount it to an nginx container so that we don't have to build the image?

Dockerfile
nsklikas, line 1: concern: this will eventually drift from the rockfile, wouldn't it be more practical to wrap the docker command with a script that builds the rock?

ui/config/useAppConfig.tsx
```diff
@@ -66,7 +66,11 @@ function AppConfigProvider({
   );
 
   useEffect(() => {
-    fetch("../api/v0/app-config", { cache: "no-store" })
+    const apiBase = process.env.NEXT_PUBLIC_API_URL ?? "";
```
nsklikas, line 69: Do we need this? This is calculated at build time so it can't be used for routing.

docker/traefik/login-ui-routes.yml
```diff
@@ -62,10 +62,14 @@ http:
...
       rule: "PathPrefix(`/ui`)"
-      service: login-ui-public-api-service
+      service: frontend-ui-service
 
   services:
     login-ui-public-api-service:
       loadBalancer:
         servers:
           - url: "http://172.17.0.1:4455"
+    frontend-ui-service:
```
nsklikas, line 72: not 100% sure: Isn't this duplicated by the changes to the docker-compose?

nsklikas (commented):
Overall lgtm, I have some minor comments.

Some extra notes/questions:
- We will need to set the CSP headers when we serve the FE from S3 from the login-ui charm. Maybe we should add them to the nginx config so that we don't forget it
- Is serving the frontend from S3 preferable to simply packing it in an nginx image and serving from there?
- When the frontend is moved to a new repo, we will need to somehow be able to fetch it so that the dev setup still works. Any idea how that will work?

ui/Dockerfile
BarcoMasile: TBD: this will be also part of the discussion with the team as we don't have an answer yet and we need to think about how to charm it properly

(btw I'm a big fan of the ngnix solution)

ui/config/useAppConfig.tsx
BarcoMasile: I'd leave it, as it's part of the agent implementation and it's coherent with the other changes

Dockerfile
BarcoMasile: I haven't given this thought, this piece of dockerfile is only for making sure the separation is successful.
After this PR we're going to create a separate repository so this will not be here anymore
Depending on how we want to take care of the separated frontend at the charm level, we can rely on the rock or on whatever we feel it's best
I think the rock is a great candidate

docker/traefik/login-ui-routes.yml
BarcoMasile: I think it relies on the nginx spun by the docker compose, as it uses it's dns basically

wood-push-melon:
One question: does the backend need to specify `AllowedCredentials` to true?

cmd/serve.go
wood-push-melon, line 145: A question (might be wrong): should we remove the `KratosPublicURL`?

BarcoMasile: I'll take note of this and check
This PR won't be merged for now, it's needed as a POC for proving the separation can happen.
Subsequent adjustments will come after we close this story

I wouldn't want to change the logic and risk breaking anything since the split is already a commitment

docker/traefik/login-ui-routes.yml
nsklikas: I meant that we define `labels` in the docker-compose that (I think) do the same thing, see https://github.com/canonical/identity-platform-login-ui/pull/970/changes#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R39:

```yaml
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.frontend-ui.rule=PathPrefix(`/ui`)"
      - "traefik.http.routers.frontend-ui.entryPoints=web"
      - "traefik.http.services.frontend-ui.loadbalancer.server.port=80"
```

BarcoMasile: so in login-ui-routes it is required to route backend APIs and the /self-service middleware to the local host (172.17.0.1:4455), since the Go backend used (especially for quick feedback loop) is not used as a container and doesn't have labels
But the labels you pointed out can be safely removed so all the networking part for this is in the yaml only.
Coming up in the next commit. Thank you

BarcoMasile:
> One question: does the backend need to specify `AllowedCredentials` to true?

If you refer to the `chi` configuration, answer is "no" in case deploy the whole system behind traefik (as this POC does) since we wouldn't have CORS issues
@wood-push-melon

go.mod
```diff
@@ -43,6 +43,7 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/leodido/go-urn v1.5.0 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
+	github.com/ory/hydra-client-go/v2 v2.2.1 // indirect
```
natalian98, line 46: do we still need that or could we stick to v26.2.0?

BarcoMasile: I think we should be able to go with the newest

PR #974, opened 2026-09-16
feat: support the RFC 8707 resource parameter

### LINKED ISSUE #966: Support RFC 8707 resource parameter
#### Context
Upstream Ory Hydra uses the non-standard `audience` parameter for API scoping and does not currently support the RFC 8707 `resource` parameter. This causes friction with strict OAuth2 clients (such as MCP) that expect standard compliance and explicitly require the `resource` parameter to bind access tokens to specific resource servers.

#### Proposal
Instead of patching upstream Hydra, we can work around this limitation directly in the login/consent server.

#### Task
- Parse the `resource` query parameter from incoming authorization requests.
- Map and append the parsed value(s) to the `audience` list when accepting the consent request with the Hydra Admin API.

#### References
* **RFC 8707:** [Resource Indicators for OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc8707)
* **MCP Protocol Issue #1614:** [Request: make `resource` parameter OPTIONAL (RFC 8707-aligned)](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1614)
* **MCP Inspector Issue #737:** [MCP Inspector OAuth Flow – Handling Unsupported resource](https://github.com/modelcontextprotocol/inspector/issues/737)

nsklikas:
## Description
Upstream Hydra scopes access tokens with the non-standard `audience` authorization request parameter and ignores the RFC 8707 `resource` parameter that strict clients (MCP) require. This PR closes the gap in the consent endpoint without patching Hydra.

- `pkg/extra`: `AcceptConsent` now grants, on top of Hydra's `requested_access_token_audience`, every `resource` value found in the query string of the original authorization URL (`OAuth2ConsentRequest.request_url`).
- Indicators are held to the rule fosite applies to the `audience` parameter (`DefaultAudienceMatchingStrategy`): the client's registered `audience` must permit them — equal scheme and host, path equal to or nested under the registered path. Deliberately stricter than fosite on opaque URIs (equal opaque part required, so one `urn:` audience does not permit every other). Hydra applies consent-granted audience unconditionally, so this is the only enforcement point.
- Indicators that are not an absolute URI, carry a fragment, or are not registered on the client are dropped and debug-logged; the consent is still accepted.
- OpenSpec change `restrict-resource-to-client-audience` (archived) with the `resource-indicator-audience` spec synced to `openspec/specs/`.

### Scope notes
- `resource` values are granted only for audiences registered on the client. Clients registered via dynamic registration without an `audience` receive no `resource`-derived audience; register the resource server on the client (`hydra create client --audience <url>`) to enable it.
- Device flow, JAR (`request`) and PAR (`request_uri`) requests are not covered: their `resource` never reaches the authorization URL Hydra stores.
- The consent is accepted with the remaining audience rather than rejected with `invalid_target` (RFC 8707 §2.1), since the request has already passed Hydra's validation by then.

## Validation
- Unit tests in `pkg/extra`: extraction and syntax boundaries (`TestResourceIndicators`), audience matching boundaries incl. `/v1` vs `/v10`, trailing slashes, scheme/host/port mismatch, `urn:` (`TestAudiencePermits`), partitioning (`TestPermittedResources`), merge/dedupe without aliasing (`TestMergeAudience`), and service-level grants with registered and unregistered clients.
- `go vet ./...` and `go test ./...` pass.
- Live proof against an isolated Hydra 25.4.0: a client registered with `audience: ["https://mcp.example.com/sse"]` sending `resource=https://mcp.example.com/sse&resource=https://other.example.com` received a JWT with `aud: ["https://mcp.example.com/sse"]` (Hydra itself reported `requested_access_token_audience: []`); the same request from a client with no registered audience yielded `aud: []`.

Closes #966


shipperizer (approved):
looks good to me

